PR #1, opened 2020-10-30
Allow components to be references

steve8708:
this to allow for cases where bundlers or node need explicit references and can't resolve dynamic imports (`import(r.component)`)

e.g. 

```ts
{ path: '/', component: () => HomePage },
{ path: '/about', component: () => import('/about') }
```

this may break the nested routing - open to feedback


ryansolid:
This should be fine I think we leave the string for lazy and we directly put the Component.. no need for the extra function wrapper. I will merge this and make the small change.

PR #2, opened 2020-11-07
Export RouteDefinition

amoutonbrady:
Exporting RouteDefinition is useful in user land when you are defining your routes in a separate file.

This would allow for something like that:

```ts
import type { RouteDefinition } from 'solid-app-router'

export const routes: RouteDefinition[] = [
  {
    path: '/',
    component: 'pages/home.tsx'
  }
]
```


ryansolid:
Thanks

PR #4, opened 2021-01-01
Adding a way to mark `Link` as external

amoutonbrady:
I took inspiration from [sapper](https://sapper.svelte.dev/docs#rel_external) router.. Might not be the best way to handle that but I figured trying to parse the `href` and handle all the case (if it starts with `/`, if it contains the same `origin`, etc..) would be much more tedious, error prone and adding more size than just allowing the end user to mark a link as external.

We do have a use case for this on the solid website where we have a bunch of links that are in an array and rendered into the navbar. Some are external, some are internal.

We could also do a `<Dynamic component={item.external ? 'a' : 'Link'} ... />` but that felt a bit much compared to a couple of line in the router

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -11,13 +11,17 @@ import {
 } from "solid-js";
 import { Show, assignProps, isServer } from "solid-js/web";
 import { RouteRecognizer, Route as RouteDef } from "./recognizer";
-import type { BaseObject, Params, QueryParams, RecognizeResults } from './recognizer';
+import type { BaseObject, Params, QueryParams, RecognizeResults } from "./recognizer";
 
 export { parseQueryString, generateQueryString } from "./recognizer";
 export type { Params } from "./recognizer";
 
 export type DataFnParams<T> = { params: Params<T>; query: QueryParams };
 export type DataFn<T = BaseObject> = (props: DataFnParams<T>) => BaseObject;
+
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah I think it's interesting in that Sapper's approach is a very HTML forward type of approach. Using rel external. This makes me want to look around a bit. The obvious way is if you include a protocol it's outside.. ie. a https:// etc.. but I want to take look around.

specious:
Upgrading solid router from 0.2.1 to 0.3.0 broke the login flow in the [littermap](https://github.com/earthstewards/littermap) app.

Bisecting the code, I found where it now [checks](https://github.com/solidjs/solid-router/commit/e31a68553807f229daad1c0bcbc7f568d2e20416#diff-692e2f97144046b5098ccb96849867854a2811345bd2ee991c0fc6d2e47f4a38R405) if an `<a>` has `rel="external"` due to regular anchor tags now also being treated as an SPA link.

Only after that I found this discussion.

For anyone who comes here wondering how to fix a similar issue, [setting](https://github.com/earthstewards/littermap/blob/e31025596a82d0e6f0a52c64229f62ed642f4ba0/src/app/ui/parts/ProfileAvatar.jsx#L33) this property on the anchor element releases it from the SPA routing logic and then it works like an normal external link.

Using solid router 0.4.2 now with no problems.

PR #14, opened 2021-05-22
Implement useHistory and useMatch

romgrk:
This adds exports `useHistory` and `useMatch`. It's annoying to have to discard one or the other of the router objects when one needs only one of the two.

This is a semver minor change.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -79,6 +79,14 @@ export function useRouter() {
   return useContext(RouterContext);
 }
 
+export function useMatch() {
+  return useContext(RouterContext)[0];
+}
+
+export function useHistory() {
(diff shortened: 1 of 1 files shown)


ryansolid:
Sorry haven't responded until now. Mostly that work on 1.0 took me away. I'm not really looking at API-level stuff here just yet. It's quite likely it will still change so I'm hesitant to really put stuff in that I might just remove immediately after. There are some big plans for the router coming down the pipeline so I appreciate the patience.

ryansolid:
Closing as we have completely refactored and added similar functionality.

romgrk:
Cool, no problem :)

PR #16, opened 2021-06-19
Fixed the issue with root path not being applied as the base to the href attribute of Link component.

Morritz:


ryansolid:
Awesome thanks.

PR #17, opened 2021-06-22
Tweaks

aminya:
I optimized the regexes used in the code:
- escapeRegex
- PATH_SEGMENT_ENCODINGS 

I also modernized some parts of the code by
- using string interpolation
- using `??` instead of `||`
- using `?.` instead of successive `&&` checks
- using `includes` instead of `indexOf()===-1`

I used my [eslint-config-atomic](https://github.com/atom-community/eslint-config-atomic) to detect and fix these issues, but I didn't commit it to the project as I was not sure if you are interested in using the plugin itself. If you are interested, let me know so I commit it. The config is [very minimal](https://github.com/atom-community/eslint-config-atomic#usage).

There were some other issues with the code, but I decided to go with this batch as the first contribution.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -142,7 +142,7 @@ export const NavLink: Component<NavLinkProps> = props => {
   return (
     <Link
       {...props}
-      classList={{ [props.activeClass || "active"]: isActive(props.href, props.exact) }}
+      classList={{ [props.activeClass ?? "active"]: isActive(props.href, props.exact) }}
     >
       {props.children}
     </Link>
@@ -188,7 +188,7 @@ function createRouter(
   );
   const current = createMemo(
     () =>
-      recognizer.recognize(root + location()) || ([] as unknown as RecognizeResults<RouteHandler>)
+      recognizer.recognize(root + location()) ?? ([] as unknown as RecognizeResults<RouteHandler>)
   );
   const data: unknown[] = [];
   const [pending, start] = useTransition();
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -142,7 +142,7 @@ export const NavLink: Component<NavLinkProps> = props => {
   return (
     <Link
       {...props}
-      classList={{ [props.activeClass ?? "active"]: isActive(props.href, props.exact) }}
+      classList={{ [props.activeClass || "active"]: isActive(props.href, props.exact) }}
     >
       {props.children}
     </Link>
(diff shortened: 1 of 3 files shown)


ryansolid:
Assuming no transpilation do you think we are safe using optional chaining and null coalescing operators?

aminya:
> Assuming no transpilation do you think we are safe using optional chaining and null coalescing operators?

Yes. `??` and `?.` guard against null/undefined. I checked these against the types.

ryansolid:
I meant for browser support. I've been avoiding these since I was worried they are too new.

aminya:
Here is the support chart for these two.

https://caniuse.com/?search=optional%20chaining
https://caniuse.com/?search=Nullish%20coalescing

If you are worried about the support, you should instead change the target in the `tsconfig`. It is very error-prone to avoid these kinds of things without a compiler watching. 

You can set this to ES2018 or something to make sure it works everywhere.
https://github.com/solidjs/solid-app-router/blob/9aaac1d33e059601f4a1a60f02f81cbd17d692a6/tsconfig.json#L4

It is also common to have two `tsconfig.json` files. One set to `esnext`, and the other one extending the modern `tsconfig` and setting an older target.

ryansolid:
I do it mostly because I dislike transpiled code. If they convert it I might never know. I'd rather something break than let bad compilation sneak in.

aminya:
> I do it mostly because I dislike transpiled code. If they convert it I might never know. I'd rather something break than let bad compilation sneak in.

We share the same feelings 😄, but that's the issue we need to deal with. 

For those who don't transpile, we need to consider the worst-case and use an old target.

For those who transpile, we need to use a modern target. I also noticed that `src` folder isn't included in the `files` entry of `package.json`. Bundlers like Parcel look for a `source` entry in `package.json`, and they use that to directly consume untranspiled code. Parcel accepts TypeScript as the source too. So, by adding `source` entry I think this issue is mostly fixed.

ryansolid:
Or we could just not adopt new syntax until it's more commonly adopted. I have a mild aversion to anything I haven't benchmarked either so I'm usually slow to adopt features.

Hmm... Thoughts about including src: 

1. Parcel is the only one that respects it I think.
2. I need a different ESM compatible way of including src for Solid's custom JSX format, the "solid" field
3. I don't like relying on TS for transformations

Mind you I guess we could be more conservative on the compilation of libraries like this since I expect people to use the "solid" field for the most part. I think while source might be useful on the main library.. even with the bloat. It has less value in libraries. I did look at the source field before introducing the "solid" field but it wasn't working for me in parcel with custom JSX transform, and the others didn't support it.

aminya:
I reverted the usage of ?? and ?.

ryansolid:
Thanks for humoring me on this.

PR #18, opened 2021-06-22
fix: add source entry to package.json

aminya:
See https://github.com/solidjs/solid-app-router/pull/17#issuecomment-865675780


ryansolid:
I don't think we are going this direction.

PR #38, opened 2021-08-18
Expose Route Props

kevineaton603:
Reasoning: https://github.com/solidjs/solid-app-router/issues/37


rturnq:
Thanks, looks like an oversight not exporting these.

PR #42, opened 2021-09-18
Replace navigate with navigatorFactory on RouterContext

svicalifornia:
This PR resolves an issue in using a `navigate` function (returned from `useNavigate`) to navigate to a path relative to the current route context.

For example, in this component:

```tsx
function MyComponent() {
  const navigate = useNavigate();

  function handleCloseButtonClick() {
    if (navigationAllowed()) {
      navigate('some/relative/path');
    }
  }

  ...
}
```

…the relative navigation will fail, because the `navigate` function does not retain the route context of the component instance. The `navigate` function invokes an internal `useRoute` function (in `src/routing.ts`) to get the current route, but the route context provider has already lost the route that was in effect when `MyComponent` was invoked and therefore returns `undefined`. Consequently the router's base route (typically `/`) is used as a fallback for path resolution, which yields an invalid path relative to the router base route, instead of the correct path relative to the component's route.

The `Link` component resolves its path relative to the route context at render time and therefore doesn't suffer the same problems as with a delayed `navigate`, but it may not satisfy all of the use cases possible by calling a `navigate` function directly, such as the conditional navigation example above.

It seems reasonable to expect that `useNavigate` should return a `navigate` function that can be invoked any time later to navigate relative to the original route context of the component from which `useNavigate` was called. Accordingly, this PR replaces the `navigate` function in the router context with a navigator factory, which returns a generated `navigate` function that references the route context currently in effect when the factory is called. This allows relative navigation to work in the above example or anywhere else where `navigate` needs to be called after the component is constructed and rendered.

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -65,7 +65,7 @@ export const useHref = (to: () => string | undefined) => {
   });
 };
 
-export const useNavigate = () => useRouter().navigate;
+export const useNavigate = () => useRouter().navigatorFactory();
 export const useLocation = () => useRouter().location;
 export const useIsRouting = () => useRouter().isRouting;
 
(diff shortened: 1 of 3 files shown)


rturnq:
This looks great. Thank you for the help!

PR #52, opened 2021-11-15
Add a new "hook" useSearchParams

rturnq:
This PR adds a new "hook" `useSearchParams` which makes getting and updating query parameters easier. `useSearchParams` returns a getter/setter tuple.

- The getter is the same object `useLocation().query` so it is a reactive proxy and not a function.
- The setter function expects a POJO containing the query parameters you want to update.
  - Updates are merged so to remove an existing query parameter, you need to set it to `''`, `udefined`, or `null`.
  - Updates use a navigation and have `scroll: false` applied by default

This PR also exports a utility function `_mergeSearchString` (underscore prefix to note instability) which powers the setter described above and can be useful for creating query strings for Links. In the future this will probably be replaced with a better API for links/navigations which allow defining hrefs as an object where you can specify query params to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,18 +168,12 @@ const params = useParams();
 const [user] = createResource(() => params.id, fetchUser);
 ```
 
-### useLocation
-
-Retrieves reactive `location` object useful for getting things like `pathname` or `query`
-
-```js
-const location = useLocation();
-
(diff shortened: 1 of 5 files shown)


ryansolid:
This all looks pretty good.

PR #53, opened 2021-11-19
Allow to pass some state between routes

SanichKotikov:
Hi,

For context https://github.com/solidjs/solid-app-router/issues/51

This is just a draft, please, review. If it's fine, I'll try to create some tests and update readme.

The main idea is able to pass some state between routes. For example:

```jsx
// route a
navigate('/b', { state: { value: 1 } });

// route b
const location = useLocation<{ value: number }>();
const valueFromPrevRoute = location.state?.value; // number | undefined
```

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -1,4 +1,4 @@
-import type { Component, JSX } from "solid-js";
+import type { Component, JSX, Accessor } from "solid-js";
 import {
   createComponent,
   createContext,
@@ -69,7 +69,7 @@ export const useHref = (to: () => string | undefined) => {
 };
 
 export const useNavigate = () => useRouter().navigatorFactory();
-export const useLocation = () => useRouter().location;
+export const useLocation = <S = unknown>() => useRouter().location as Location<S>;
 export const useIsRouting = () => useRouter().isRouting;
 
 export const useMatch = (path: () => string) => {
@@ -195,7 +195,7 @@ export function getRouteMatches(branches: Branch[], location: string): RouteMatc
   return [];
 }
 
-export function createLocation(path: () => string): Location {
+export function createLocation(path: Accessor<string>, state: Accessor<any>): Location {
   const origin = new URL("http://sar");
   const url = createMemo<URL>(
     prev => {
@@ -216,7 +216,6 @@ export function createLocation(path: () => string): Location {
   const pathname = createMemo(() => url().pathname);
   const search = createMemo(() => url().search.slice(1));
   const hash = createMemo(() => url().hash.slice(1));
-  const state = createMemo(() => null);
   const key = createMemo(() => "");
 
   return {
@@ -230,7 +229,7 @@ export function createLocation(path: () => string): Location {
       return hash();
     },
     get state() {
-      return state();
+      return state() || null;
     },
     get key() {
       return key();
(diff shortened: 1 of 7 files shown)


src/routing.ts
```diff
@@ -318,11 +318,15 @@ export function createRouterContext(
...
           }
-          setSource({ value: resolvedTo, replace, scroll });
+          setSource({ value: resolvedTo, replace, scroll, state });
         } else {
-          const len = referrers.push({ value: current, replace, scroll });
+          const len = referrers.push({ value: current, replace, scroll, state });
           start(
-            () => setReference(resolvedTo),
+            () => {
+              // TODO: not sure about batch()
```
SanichKotikov, line 326: Should I wrap these lines in batch() or `start()` already does it?

src/routing.ts
```diff
@@ -348,7 +352,8 @@ export function createRouterContext(
         setSource({
           value: next,
           replace: first.replace,
-          scroll: first.scroll
+          scroll: first.scroll,
+          state: first.state
```
rturnq, line 356: I'm not sure this is exactly what we want here, but maybe it is. The router batches synchronous navigations - that is to say if you navigate with a link or `useNavigate`, if another navigation occurs during rendering from something like a `<Navigate />` component we treat them as one navigation. So here `first` refers to the original navigation event which would mean if a redirect provided a state as well, that would be lost. The alternatives are pass in the state to this function and always use it or maybe use the last one defined.

src/routing.ts
```diff
@@ -230,7 +229,7 @@ export function createLocation(path: () => string): Location {
       return hash();
     },
     get state() {
-      return state();
+      return state() || null;
```
rturnq, line 232: Since state could be anything (including things like `0`, `false`, etc.) we should be a little stricter here and only return `null` if `state() == null` (ie. `null` or `undefined`).

rturnq (changes requested):
Thanks for this PR, this is on my TODO list.

src/routing.ts
SanichKotikov: If we will use `useNavigate` with custom state, and this redirects with `<Navigate />`, we will lost the state and got undefined. `react-router` and native `.pushState` works in the same way. So, I suppose it works as expected right now, or I messed something.

src/routing.ts
SanichKotikov: We can return just `state()`, this returns `undefined` if there is no state.

rturnq:
I made a couple additions
- Added `state` prop to the `Navigate` component
- Update `createIntegration` to allow the path integration to return the `history.state` value which allows the state to be pulled from history when you go back and forward.

src/routing.ts
rturnq: I agree, if a redirect occurs, then the state should be whatever was used on that redirect navigation which likely will be undefined. The only problem as it stands is there is a mismatch between what gets set in the `state` signal and what gets passed to the integration. The state signal will get the state from the last navigation event and the integration will be sent the first event's state.

Here is an example: https://codesandbox.io/s/solid-app-router-pr53-1-0c3ek
- Click `foo` link, notice state is `{ from: 'home' }
- Click `baz` link, notice state is `undefined`
- Click browser back, notice state is `{ from: 'home' }` (expected)
- Click browser forward, notice state is `{ from: 'foo' }` (unexpected, should be `undefined`)

rturnq: My latest commits makes this work like I would expect. Example (use the same steps as above): https://codesandbox.io/s/solid-app-router-pr53-2-un0bi

SanichKotikov:
Nice updates. Thx!
Is there something else we should add or is everything ok?

rturnq:
Looks good.

PR #56, opened 2021-12-01
Setup development build

rturnq:
This PR adds a development build so we can include development only code which will be conditional on the "_SOLID_DEV_" literal. This code will be removed by rollup in the same way as Solid proper and exported on the "development" condition.

This PR also adds a utility `warnOnce` which logs to console one time per message.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "solid-app-router",
       "version": "0.1.13",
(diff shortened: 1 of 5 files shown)


ryansolid:
One thought is this dev build will only work when people aren't doing SSR. Or rather not using vite because "solid" will resolve first. I think we might need to nest if we want this to work in Vite.. however, that will require doing the replace in an `.jsx` export so can't build it the same way.

PR #59, opened 2021-12-11
Allow listing for memo objects

Schniz:
This is handy when someone wants to `JSON.stringify(params)`, but with extra steps (since the Proxy does not implement `toJSON`):

```ts
function DebugRoute() {
  const params = useParams();
  const stringifiedParams = Object.fromEntries(
    Object.getOwnPropertyNames(params).map((key) => [key, value])
  );
  return <pre>{JSON.stringify(stringifiedParams, null, 2)}</pre>;
}
```

I also wondered if I should just implement `toJSON`, but it feels too specialized for this use case.

Please let me know if there is another way of doing so, without this contribution. I would be happy to use "the standard" way instead of adding more lines of code :smiley:

Thanks for an amazing product by the way! It's really exciting.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -112,6 +112,12 @@ export function createMemoObject<T extends object>(fn: () => T): T {
           });
         return memo();
       },
+      getOwnPropertyDescriptor() {
+        return {
+          enumerable: true,
+          configurable: true
+        };
+      },
       ownKeys() {
         return Reflect.ownKeys(fn());
       }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -111,6 +111,9 @@ export function createMemoObject<T extends object>(fn: () => T): T {
             return p;
           });
         return memo();
+      },
+      ownKeys() {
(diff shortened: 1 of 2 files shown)


rturnq:
Thanks for this. It looks like if we also add this to allow the keys to be enumerable, we can also support `Object.keys` as well as `JSON.stringify`. For example: https://codesandbox.io/s/solid-app-router-pr59-f2sw4
```js
getOwnPropertyDescriptor() {
  return {
    enumerable: true,
    configurable: true
  };
}
```

Schniz:
Great idea, pushed a fix with tests 😃

Schniz:
oh no, now I see that prettier made a lot of changes in the test file

Schniz:
@rturnq would you want me to revert the stylistic changes done by prettier and keep my changes local? 😄

rturnq:
The prettier changes are fine. Looks like I need to prettify everything since it clearly wasn't. Thanks for the PR.

PR #68, opened 2022-01-05
Updated createMemoObject

high1:
- do not rely on the return value of runWithUpdates
- updated typings and removed casting
- typings for helpers.ts


rturnq:
Thanks again.

PR #69, opened 2022-01-07
Negative numbers in useData for counting from top

edemaine:
As [discussed in #website](https://discord.com/channels/722131463138705510/722167449251741798/929055041766174770), this adds functionality to `useData` so that negative numbers count from the root instead of the child.

I've tested this via `npm link` in `solid-website`, but it'd be great to add local tests for `useData` (there are currently none). Would someone else be willing to do that, here or in another PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ return <div classList={{ "grey-out": isRouting() }}>
 
 ### useData
 
-Retrieves the return value from the data function. You can access parent data by passing a number to indicate ancestor. No argument is the route's own data, `1` is the immediate parent, `2` is the parent's parent, and so on.
+Retrieves the return value from the data function. You can access parent data by passing a number to indicate ancestor. No argument (or `0`) is the route's own data, `1` is the immediate parent, `2` is the parent's parent, and so on. Alternatively, `-1` is the top-level route's data, `-2` is the immediate child, etc.
 
 ```js
 const user = useData();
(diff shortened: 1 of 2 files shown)


ryansolid:
I see makes sense. But also adds a nice chunk of new code. Is there just a better way to access parent data I wonder.  I'm open to suggestions. I might want to look at what Remix is doing.

Just went this way for lack of better idea, but now we are building a system on it. I don't think it is a bad one per se. But before we go here I wonder if we can't just look at this more holistically.

edemaine:
The code would definitely be simpler if there were an array of ancestors available. I'm not sure how feasible that is.

I'd probably find the API more intuitive if the order was flipped: 0 means root, -1 means deepest child.  But that's probably a needlessly breaking change.

In any case, I think we need both indexing from the top (to get at the root route) and indexing from the bottom (current route). If not negative numbers, then two different API calls...

edemaine:
FYI, solid-site is currently depending on this for dark mode (via [this already merged PR](https://github.com/solidjs/solid-site/pull/221)).  Perhaps it would be worth merging this in as a natural extension of the current numeric model, and introducing a new/breaking model for `useData` in the future? Or you're worried about code bloat that we can't remove later for backward compatibility sake?

I'll open a separate issue about the new/breaking model.

ryansolid:
No longer direction we are taking. Thanks for the PR and the ensuing discussion as it led to a solution I think we are all happier with.

PR #98, opened 2022-04-08
Docs Rewrite

Jutanium:
So far, I left the Router Primitives section the same, but I provided a full guide to using the Router. I did move the Config Based Routing section until after the explanations because the JSX-based config is easier for explaining the concepts.


README.md
```diff
@@ -30,58 +52,306 @@ render(
...
+
+You can then access that `id` from within a route component with `useParams`:
+
+```jsx
+//async fetching function
+import { fetchUser } ...
+
+export default function User () {
+
+  const params = useParams;
```
pawelblaszczyk5, line 207: Shouldn't useParams be invoked? :)

PR #105, opened 2022-04-21
Fix self-closing Route tags

rpivo:
Was copying some of this code to a project and noticed my route wasn't working because it was missing the slash


ryansolid:
Thank you

PR #106, opened 2022-04-26
Fix #96 - support replace with hashIntegration

### LINKED ISSUE #96: navigate(path, {replace: true}) with hashIntegration
We're using hashIntegration in our app, but the {replace: true} option in the navigate function doesn't seem to "replace" the current path in window.history. Is this a work-in-progress?

There doesn't seem to be an equivalent for this [code](https://github.com/solidjs/solid-app-router/blob/99e977b970a31cba31b41e9698c78143afeb43fd/src/integration.ts#L86-L90) in [here](https://github.com/solidjs/solid-app-router/blob/99e977b970a31cba31b41e9698c78143afeb43fd/src/integration.ts#L104-L107)

Brendan-csel:
Not 100% sure if the `(hashIndex === 0 ? "" : "#")` check is required or if we can always simply always prepend '#'.  When testing here `value` never had the leading '#' - but others may use differently?

### DIFF
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -110,12 +110,12 @@ export function hashIntegration() {
   return createIntegration(
     () => window.location.hash.slice(1),
     ({ value, replace, scroll, state }) => {
-      const hashIndex = value.indexOf("#");
       if (replace) {
-        window.history.replaceState(state, "", (hashIndex === 0 ? "" : "#") + value);
+        window.history.replaceState(state, "", "#" + value);
       } else {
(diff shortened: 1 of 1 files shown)


rturnq:
Thanks for the PR.

All the paths should be normalized by the time they get to the integration so there should never be a leading `#`. I think we can remove that.

Brendan-csel:
Sweet - I've cleaned that up

rturnq:
Thank you.

PR #109, opened 2022-05-05
fix #108 useSearchParams bug

### LINKED ISSUE #108: useSearchParams is changing twice after go back in browser history
After setting a new search string, go back in the browser history and try to set it again, the search string will be changed twice. The first change will be the value that I'm trying to set and the second change will be the previous value.

I think the problem is because the reactive location.search is trying to re-execute. The solution would be wrap mergeSearchString with an untrack?

https://github.com/solidjs/solid-app-router/blob/92131af5bebce2f3993283349dcddf72f9506d75/src/routing.ts#L94-L97

mishimalisa:
fixes https://github.com/solidjs/solid-app-router/issues/108

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -92,7 +92,7 @@ export const useSearchParams = <T extends Params>(): [
   const location = useLocation();
   const navigate = useNavigate();
   const setSearchParams = (params: SetParams, options?: Partial<NavigateOptions>) => {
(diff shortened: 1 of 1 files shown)


rturnq:
Thanks, I'll merge this since it will improve the case where you call it from an effect, and make another broader fix to work around the underlying bug.

PR #126, opened 2022-06-17
fix typings in Route

chaurka:
A recent [commit](https://github.com/solidjs/solid-app-router/commit/9b3348558465058f6d0a7a1740076c1ecd622ed8) broke the typings because of removed `as unknown as JSX.Element` cast.


ryansolid:
Thanks.

PR #138, opened 2022-07-15
Update useRouteData return type annotation

andrewrosss:
Saw this on stream.

It was mentioned that the ergonomics of having to use the built-in `ReturnType` type when calling `useRouteData` could be improved. Maybe something like this could be useful?

With this change you can just pass `typeof myFunc`:

```ts
function userData() {
  const [user] = createResource(async () => ({ id: 123, name: "hello" }));
  return user;
}

const user = useRouteData<typeof userData>();
//    ^? {id: number, name: string} | undefined
```

while still supporting the old type annotation, i.e.:

```ts
const user = useRouteData<ReturnType<typeof userData>>();
//    ^? {id: number, name: string} | undefined
```

Of course, if this isn't desirable or there are plans for something else, feel free to disregard/close. Great work either way! :)

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -1,4 +1,19 @@
-import { Accessor, Component, JSX, createResource } from "solid-js";
+import type { Component, JSX, Accessor } from "solid-js";
+import {
+  createComponent,
+  createContext,
+  createMemo,
+  createRenderEffect,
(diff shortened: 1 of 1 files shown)


ryansolid:
Thank you.

PR #139, opened 2022-07-18
Merge `classList` props on NavLink

thibautRe:
This PR adds support for the `classList` prop on the `NavLink` Component.

Before this PR, this pattern would be problematic:

```tsx
<NavLink classList={{ [myClass]: true }} />
```

The NavLink would just entirely void the prop and overwrite it with its internal classes, leading to confusion.


ryansolid:
Thank you.

PR #169, opened 2022-09-20
Set isRouting only when routing

### LINKED ISSUE #166: useIsRouting resolves to a true value for any transition.
### Describe the bug

`useIsRouting()` returns `true` for any transition happening within the app, not necessarily for anything originating in the navigation. This makes implementing a global route loading state indicator impossible, as it will show false-positives.

https://stackblitz.com/edit/vitejs-vite-czcxit?file=src/App.tsx

1. Navigate to the `Other` route, see that the `Router state` changes to `routing` during the transition.
2. Click on the `re-roll` button. The `Router state` changes to `routing`

### Expected behavior

The `routing state` should remain `idle` as no navigation is happening.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser Chrome
- Version: 1.5

### Additional context

_No response_

jeski-bright:
Create a new signal to indicate whether the router is navigating or not, do not rely on the `transitioning` state.

Fixes #166

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -9,7 +9,7 @@ import {
   onCleanup,
   untrack,
   useContext,
-  useTransition,
+  startTransition,
   resetErrorBoundaries
 } from "solid-js";
 import { isServer, delegateEvents } from "solid-js/web";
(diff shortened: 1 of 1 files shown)


ryansolid:
This isn't a bad approach.. It's always async but maybe that is ok. Basically it will always flicker. I think that is passable for now but I want to think of if we can do better.

PR #175, opened 2022-09-30
Named Routes

Supertigerr:
## Why
Named routes can be really helpful when doing something complex.. For example if you have a drawer pane and a main content pane, you can add custom logic to handle that. It also saves you from component remounts.

If any modifications are needed, let me know/edit it :)

## Example
```tsx
import { Route, Routes, useRouteName, Link } from '@solidjs/router';

  return (
    <>
      <h1>My Site with Lots of Pages</h1>
      <Link href='/'>Home</Link>
      <Link href='/users'>Users</Link>
      <Routes>
        <Route path="/" component={Page}>
          <Route name='home' path='/' />
          <Route name='users' path='/users'  />
        </Route>
      </Routes>
    </>
  );
};

const Page = () => {
  const name = useRouteName()
  createEffect(() => {
    console.log(name());
  })
  return <div>{Math.random()} - {name()}</div>
}
```

### DIFF
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -150,6 +150,7 @@ export const useRoutes = (routes: RouteDefinition | RouteDefinition[], base?: st
 
 export type RouteProps = {
   path: string | string[];
+  name?: string;
   children?: JSX.Element;
   data?: RouteDataFunc;
 } & (
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,7 +9,8 @@ export {
   useNavigate,
   useParams,
   useResolvedPath,
-  useSearchParams
+  useSearchParams,
+  useRouteName
 } from "./routing";
 export { mergeSearchString as _mergeSearchString } from "./utils";
 export type {
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -83,6 +83,7 @@ export const useMatch = (path: () => string) => {
   return createMemo(() => matcher()(location.pathname));
 };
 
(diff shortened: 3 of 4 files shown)


ryansolid:
To clarify the name isn't doing anything here other than serving as additional metadata. This isn't a way to navigate using the name or anything like that?

Supertigerr:
Sorry for my bad explanation. 
My current problem is that, I have routes like this:
```tsx
<Routes>
     <Route path='app/' component={<App/>}  />
     <Route path='app/users' component={<App routeName="users" />}  />
 <Routes>
 ```
Both routes take you to the same page. But, they have different drawer and content page. Other than that, other stuff in the routes are the same. 

The problem with this is that, the whole page gets re-rendered when switching to these 2 routes instead of ONLY the drawer and the content page. 

I could do something like this instead:
```tsx
<Routes>
     <Route path='app/' component={<App/>} >
         <Route path='/' />
         <Route path='/users' />
     </Route>
 <Routes>
 ```
This will prevent re-rendering. But the problem is, now I cant add the `routeName` prop like I did in the first example in this comment. So, there is no way to change the drawer and the content page accordingly.

You may say I could use `Outlets` But That will only allow me to change ONE thing and not two (drawer and the content page)

I hope this was clear 😅

Supertigerr:
Unfortunately, I see what you mean. Adding proper named routes will require significant changes with this router. I made my own router package for this.

PR #178, opened 2022-10-05
Update redirect example in readme.MD

ivancuric:
The `<Navigate>` component isn't supposed to be a child of a `Route`, but declared on the `element` prop to work.


ryansolid:
Thanks.

PR #179, opened 2022-10-08
Streamline anchors

ryansolid:
It's been bugging me ever since working on SolidStart where we have both `@solidjs/router` and `@solidjs/meta` that there is a conflict with them both having `<Link>` components. And the truth of the matter is Solid Meta sorta takes priority because it is a wrapper over the `<link>` element.

Solid Router already diverted from React Router by using the `href` JSX attribute and we recently did work to make our anchors work even when not hydrated. So instead of having `<Link>` and `<NavLink>` we just have `<A>` (For the next while we still export `<Link>`, `<NavLink>` for backwards compat). This follows the pattern we've been using for the others allow enhancement. Not wanting to wrestle with having two components all the functionality of `<NavLink>` is built into `<A>`. 

At the same time while we need to intercept all anchor clicks I realize that our automatic opt-in of `<a>` is awkward. So with the next version we will go back to `<a>` just being normal and if you want Solid's router to do its thing use `<A>`. It still works without hydration but we now can leave your normal anchors alone.  Technically this is being handled by an attribute ("link") that is being added to the underlying `<a>` so you can still generate this without the Component but this just aligns it.

All in all this is a departure but it feels more like Solid to go this way and the right balance of power while getting out of your way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,11 @@ export default function App() {
 
 ## Create Links to Your Routes
 
-Use the `Link` component to create an anchor tag that takes you to a route:
+Use the `A` component to create an anchor tag that takes you to a route:
 
 ```jsx
 import { lazy } from "solid-js";
-import { Routes, Route, Link } from "@solidjs/router"
+import { Routes, Route, A } from "@solidjs/router"
 const Users = lazy(() => import("./pages/Users"));
 const Home = lazy(() => import("./pages/Home"));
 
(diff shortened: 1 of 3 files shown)


nksaraf:
Should we also consider changing the `end` prop. It makes very little sense. Maybe something like `exact` or `equal` or otherwise something like `activeWhen: 'exact' | 'prefix'`. Just ideas that are better than `end`. It should atleast indicate that its pertinent to the active status.

ryansolid:
I don't think `exact` or `equal` are better. Not any more descriptive. And like the boolean. Can't think of a better descriptive name at the moment though.

PR #186, opened 2022-10-18
Support for blocking navigation (useBeforeLeave)

Brendan-csel:
Adds this...

### useBeforeLeave

`useBeforeLeave` takes a function that will be called prior to leaving a route.  The function will be called with:

- from (_Location_): current location (before change).
- to (_string | number_}: path passed to `navigate`.
- options (_NavigateOptions_}: options passed to `navigate`.
- preventDefault (_void function_): call to block the route change.
- defaultPrevented (_readonly boolean_): true if any previously called leave handlers called preventDefault().
- retry (_void function_, _force?: boolean_ ): call to retry the same navigation, perhaps after confirming with the user. Pass `true` to skip running the leave handlers again (ie force navigate without confirming).

Example usage:
```js
useBeforeLeave((e: BeforeLeaveEventArgs) => {
  if (form.isDirty && !e.defaultPrevented) {
    // preventDefault to block immediately and prompt user async
    e.preventDefault();
    setTimeout(() => {
      if (window.confirm("Discard unsaved changes - are you sure?")) {
        // user wants to proceed anyway so retry with force=true
        e.retry(true); 
      }
    }, 100);
  }
});
```

## NOTE LIMITATIONS
This PR only adds support to block navigations originating from the router (eg `<Link>`, `useNavigate()`).

Blocking navigations originating in the browser (back/forward, plain anchors) can also be done and requires either a custom path/hash integration, or `solid-router` to extend its built-in integrations. [This branch](https://github.com/cselnz/solid-router/tree/publish) includes this addtional behavior.

Finally, users may wish to block full page document navigations (ie using window `beforeunload` event) but that should be handled in application code directly (outside the scope of the router).

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,7 +544,9 @@ Used to define routes via a config object instead of JSX. See [Config Based Rout
 
 `useBeforeLeave` takes a function that will be called prior to leaving a route.  The function will be called with:
 
-- to (_path: string |number_, _options: NavigateOptions_}: same as passed to the `navigate` causing the route change.
+- from (_string_): path prior to `navigate`.
+- to (_string | number_}: path passed to `navigate`.
+- options (_NavigateOptions_}: options passed to `navigate`.
 - preventDefault (_void function_): call to block the route change.
 - defaultPrevented (_readonly boolean_): true if any previously called leave handlers called preventDefault().
 - retry (_void function_, _force?: boolean_ ): call to retry the same navigation, perhaps after confirming with the user. Pass `true` to skip running the leave handlers again (ie force navigate without confirming).
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ It supports all of Solid's SSR methods and has Solid's transitions baked in, so 
   - [useRouteData](#useroutedata)
   - [useMatch](#usematch)
   - [useRoutes](#useroutes)
+  - [useBeforeLeave](#usebeforeleave)
 
 ## Getting Started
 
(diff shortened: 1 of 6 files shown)


Brendan-csel:
Damn - for some reason the changed files are showing changes in another recent commit too `components.tsx`.  Please ignore that for now.

Brendan-csel:
Hmm - I wonder if there would ever be a time when the app would want to retry the navigate - but not _force_ it.  Example could be if prompting the user with "Would you like to save your changes before leaving", then attempt save, then retry (with all checks running again).

Perhaps instead of `e.forceRetry()` it should be `e.retry({force: true})` or just `e.retry(true)` ??

Brendan-csel:
Have added "from" path to event args as it proved useful in real-world testing.

Brendan-csel:
Hmm, I think there is an issue here requiring a bit of a code restructure.  I set up the leave handlers globally - but they should be per <Router>.  I'll work on that.

Brendan-csel:
Okay the latest commit makes the BeforeLeave subscribers specific to the Router instance.

BeforeLeave management can be created by either:
1. the Router - which means it only supports blocking router initiated navigations.
2. the integration - which means browser initiated navigations can also be blocked (back/forward, etc).

Note that 2 still requires a change to the integrations which I haven't included in this PR - but do have in my [publish branch](https://github.com/cselnz/solid-router/tree/publish) (and are using in production O_O!!)

filipjnc (approved):

ryansolid:
This looks really good, and it taps in before Transitions so I think we are good here. Love it.

PR #189, opened 2022-10-21
Fix #176 (2nd issue) - loosing trailing slash on query

### LINKED ISSUE #176: query param auto remove special chracter
### Describe the bug

`domain:verification?token=U2FsdGVkX1+Pn+CqDMQBEdFWaMVnn/sLZBPTE2V2weZxTGYMa+njTIHBBElcpgZhOyhCKjKGM16qW71gzDTu2w==&email=example@gmail.com`

the searchParams.token itself will auto remove `+` and other special characters.

Brendan-csel:
This `normalize` utility is passed paths that _can_ include a query string eg `/login?ReturnTo=/` and was stripping off the trailing slash.

It is probably still intended to strip trailing slashes from the "path" portion.  So this PR takes the approach of splitting on any "?", trimming the first item then joining the items again with "?".

Hopefully helpful but feel free to discard if not. :)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,7 +5,9 @@ const hasSchemeRegex = /^(?:[a-z0-9]+:)?\/\//i;
 const trimPathRegex = /^\/+|\/+$/g;
 
 function normalize(path: string, omitSlash: boolean = false) {
-  const s = path.replace(trimPathRegex, "");
+  const p = path.split("?");
+  p[0] = p[0].replace(trimPathRegex, "");
+  const s = p.join("?");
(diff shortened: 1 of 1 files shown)


Brendan-csel:
This solution is wrong.  

The issue is actually caused by a different bug in the router...
- the router decodes the URL from `/login?returnTo=%2F` to become `/login?ReturnTo=/`
- the router then strips the trailing / 
- BUT THEN the router and writes the url **with decoded query string** back into the address bar - including characters that shouldn't be allowed.

Issue #112, opened 2022-05-16
Strange removal of spaces in href attribute of <Link/>


titoBouzout:
Stumble upon this curious case
```js
import { Link, useParams } from 'solid-app-router'

<Link href="/blog/bla bla bla"/>

// becomes 

<a href="/blog/blablabla"/>
```

I know folder/file/path names should be escaped, but in any case this shouldn't be happening. 

Found out is related to this change, the change on the regular expression at line 5  https://github.com/solidjs/solid-app-router/commit/6ce024ed88382d23a5fc6b83c80fa593d5f7b409

I'm not completely sure what the removal of the space is trying to achieve here.

A question if you don't mind, why the path doesn't gets escaped automatically?

Thanks!

rturnq:
Looks like that string removal in the regex was inadvertent. This was a good find as we weren't properly handling spaces and other encoded characters in the path, search query and hash.

titoBouzout:
Thank you, I'm a bit sceptical about the url decoding(like in decoding something you didn't encode yourself). I will check if a very complex url works as expected whenever I have a chance, and report back.  Thanks again

PR #190, opened 2022-10-22
Fix #176 - url was being decoded, trimmed, and written back unencoded

### LINKED ISSUE #176: query param auto remove special chracter

Brendan-csel:
Fixes #176.

It looks like this issue was introduced by [this commit](https://github.com/solidjs/solid-router/commit/13d33e7fb943b9d5255ebc96fbb3e58bc5da1b4a) back in May when fixing issue #112.

The problem was decoded (invalid) URLs were being written back into the browser address bar (potentially after having trailing characters removed from query string).

I've also tested the original issue #112 with this change in place and it still works correctly....
 `<Link href="/bla bla bla">Blah</Link>`...
...is correctly matched against route `{ path: '/bla bla bla', component: Blah }`...
... and shows `/bla%20bla%20bla` in the address bar.

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -451,7 +451,7 @@ export function createRouterContext(
       )
         return;
 
-      const to = parsePath(pathname + urlDecode(url.search, true) + urlDecode(url.hash));
+      const to = parsePath(url.pathname + url.search + url.hash);
       const state = a.getAttribute("state");
 
       evt.preventDefault();


ryansolid:
Ok this is still an improvement. So lets move forward with this for now.

PR #191, opened 2022-10-22
Fix router tests

Brendan-csel:
This resolves two issues:
1. Jest apparently now requires `jest-environment-jsdom` to be installed independantly (since v29)
2. Most of the tests in `router.spec.ts` were being skipped - apparently due to a couple of tests being added as `test.only()` instead of just `test()`.

Caveat - I know nothing about Jest.  This PR is from a process of elimination (ie commenting out bits until it worked then adding them back in).

Before...
image

After...
image

Note: the remaining skipped test explicitly calls `test.skip(...)`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,67 +9,94 @@
       "version": "0.4.3",
       "license": "MIT",
       "devDependencies": {
-        "@babel/core": "^7.15.8",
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks

PR #192, opened 2022-10-22
Fix #184 - Link not matching trailing /

### LINKED ISSUE #184: When hosted under a subdirectory the root (`/`) link is not highlighted when there is a trailing `/`
### Describe the bug

I have a solidJS app hosted under `domain.com/solidjs/`.

My Router is setup with a base property: `<Router base="solidjs">`
And my vite config has `base: "/solidjs/";`

Everything seems to be working correctly, except when navigating to my URL or refreshing the page on the root path, even though the home route is correctly rendered, the following link is not given an active status
```
<A class="nav-link" href="/" end={true}>
  Home
</A>
```
This is not an issue for any of my other page links (even those that also have `end={true}`)

After investigating it seems to be entirely due to whether there is a trailing slash on my root page's URL

`domain.com/solidjs` - works correctly and the `/` link is set to active
`domain.com/solidjs/` - does not set the `/` link to active

I can tweak my nginx config to not auto-add the trailing slash, but as this is not a problem with either the react or vue routers, so I thought it worth raising.

Many Thanks

https://izrite.com/solid-explore/

Go to `https://izrite.com/solid-explore/` - see the Home link *is not* active
Go to `https://izrite.com/solid-explore` - see the Home link *is* active

### Expected behavior

I would expect Home <A> link to be set to active regardless of whether there is a trailing slash

### Screenshots or Videos

_No response_

### Platform

*

### Additional context

_No response_

Brendan-csel:
Fix #184 
Issue summary : `<A>` was not showing as active for `domain.com/solidjs/` due to trailing slash.

This PR exposes the `normalize()` utility function as `normalizePath()` so it can be used by the `<A>` component when comparing active paths.

### DIFF
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -209,7 +209,7 @@ export function A(props: AnchorProps) {
     if (to_ === undefined) return false;
     const path = normalizePath(to_.split(/[?#]/, 1)[0]).toLowerCase();
     const loc = normalizePath(location.pathname).toLowerCase();
-return props.end ? path === loc : loc.startsWith(path);
+    return props.end ? path === loc : loc.startsWith(path);
   });
 
   return (
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -27,7 +27,7 @@ import type {
   RouteDefinition,
   RouterIntegration
 } from "./types";
-import { joinPaths } from "./utils";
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks

PR #194, opened 2022-10-23
Fix #130 scroll to hash not always working

### LINKED ISSUE #130: Scroll to section with id from a hashed url
### Describe the bug

It happened to me like in the issue 83 report https://github.com/solidjs/solid-app-router/issues/83 that luckily I solved it. Since I can scroll from the same page to the section with id. Eg `<section id="contact">...</section>` using Link or NavLink.
But the problem is when I load the page from a url in a new tab, for example localhost:3000/#contact or localhost:3000#contact, the scroll is always positioned at the top.
I appreciate if you can tell me if I'm doing something wrong.

https://digichanges.com/#about

If you click on the links, you will see that it scrolls correctly

If you load the url in the browser you will see that it does not scroll to the corresponding section https://digichanges.com/#about

### Expected behavior

I would expect that when accessing the site, it would scroll to the section identified in the url after the #

### Screenshots or Videos

_No response_

### Platform

- OS: [macOS, Linux]
- Browser: [Chrome, Safari]

### Additional context

Without the router it worked fine.

Brendan-csel:
Not sure if you want to go here but...

If `scrollToHash` can't immediately find the element to scroll to, this PR sets up a `MutationObserver` to wait for the element (up to 100ms) and then scrolls to it.

Note, for performance it also uses `getElementById` instead of `querySelector` - I think that is okay for browser hashes??

### DIFF
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -14,22 +14,33 @@ function intercept<T>(
   return [get ? () => get(value()) : value, set ? (v: T) => setValue(set(v)) : setValue];
 }
 
-function querySelector<T extends Element>(selector: string) {
-  // Guard against selector being an invalid CSS selector
-  try {
-    return document.querySelector<T>(selector);
-  } catch (e) {
-    return null;
-  }
+function waitForElement(id: string, timeout: number) {
+  return new Promise<Element | null>(resolve => {
+    const el = document.getElementById(id);
+    if (el) {
+      resolve(el);
+    } else {
+      const observer = new MutationObserver(mutations => {
+        const el = document.getElementById(id);
+        if (el) {
+          resolve(el);
+          observer.disconnect();
+        }
+      });
+      observer.observe(document.body, { childList: true, subtree: true });
+      setTimeout(() => {
+        observer.disconnect();
+        resolve(null);
+      }, timeout);
+    }
+  });
 }
 
 function scrollToHash(hash: string, fallbackTop?: boolean) {
-  const el = querySelector(`#${hash}`);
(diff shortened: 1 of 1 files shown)


RAnders00:
Just a suggestion: I think it would be benefitial to have this behaviour documented - that it exists in the first place, and secondly, that the limit is 100ms.

ryansolid:
Hmm.. mutation observer can also have other adverse impact. Im gathering this came up because something wasn't rendered in time but 100ms seems arbitrary. Since Solid renders synchronous Id think like next frame would be sufficient. If not I think any heuristic is arbitrary and needs so thought.

Brendan-csel:
The first thing I tried was just a simple setTimeout but even with that set to 2ms the component hadn't rendered - and I just figured out why!!

The route component is `lazy` - so it has to load the code before it can render.  Therefore, the element we want to scroll to doesn't exist for some unknown time.  I was scratching my head looking for a data fetch or some other suspense thing, but it was simply delayed waiting for the code to load.

I appreciate the MutationObserver/timeout approach is arbitrary (and hacky) - but the current approach just fails to scroll every time so I though this might be a lesser evil as is would at least wait "a little longer" to find the element to scroll to.

Brendan-csel:
After a bit of research, I'm not sure the "other routers" do anything for this beyond what Solid Router was already doing.

Perhaps the solution here is "won't fix" - with a suggestion that scrolling to any hash id be done in the app component (when it eventually renders).

Brendan-csel:
Closing this to reduce noise.

Issue #197, opened 2022-11-10
Parent route not rendered if children is empty


bavrr:
### Describe the bug

When routes is initialized with useRoutes() and a route with an empty children exist, the parent route is not render without any infromation

### Your Example Website or App

https://codesandbox.io/s/peaceful-bassi-33scg7?file=/src/main.tsx

### Steps to Reproduce the Bug or Issue

1. Go to [https://33scg7.sse.codesandbox.io](url) - see MainPage not rendered
2. Go to [https://33scg7.sse.codesandbox.io/jop](url) - see AnotherPage is rendered

### Expected behavior

Render parent route or throw error with information.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

ryansolid:
fixed in 0.5.1

PR #198, opened 2022-11-11
Fix issue: #197

bavrr:
Fix issue [#197 ](https://github.com/solidjs/solid-router/issues/197).
Issue summary : Parent rout was not rendering without error or any information if children is empty.

This PR add condition for children routes.  If children are empty parent route will render anyway.


ryansolid:
Thanks

PR #206, opened 2022-11-21
Gurad Added with redirect also active link fix

puneetxp:
sorry last time i sync without thinking about warning.

Here is again

### Guard

When notLogin(in below) is true that mean it will not do anything but if it is string it will redirect to the string that we get in return.

```js
const notLogin: Accessor<string | true> = createMemo(() => {
  if (unwrap(LoginService.get().login) != false) { return '/' }
  return true;
});
  return (
    <>
      <Router>
        <Routes>
          <Route path='/' component={Public_Layout}>
            <Route path='' component={HomePage} />
            <Route path='login/' guard={notLogin} component={Login_Page} />
            <Route path='register/' guard={notLogin} component={RegisterPage} />
          </Route>
        </Routes>
      </Router>
    </>
  );
};
```

Also there is issue i think

```js const App: Component = () => {
  return (
    <>
      <Router>
        <Routes>
          <Route path='/' component={Public_Layout}>
            <Route path='' component={HomePage} />//this came activeclass for login path
           <Route path='login/' guard={notLogin} component={Login_Page} />//this need only activeclass
            <Route path='register/' guard={notLogin} component={RegisterPage} />
          </Route>
        </Routes>
      </Router>
    </>
  );
};
```

for resolve this issue.
change

```js
    return props ? path === loc : loc.startsWith(path);
```
to  2 options
first 
```js
    return props?.children ? path === loc : loc.startsWith(path);
```

second that is kind of cutting thing I think so. 
```js
    return path === loc;
```

i don't found any use of it but may be there are uses that i don't know about it.

Also You are free for any edit and I feel this feature is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,6 +498,33 @@ return (
     <button onClick={() => setSearchParams({ page: searchParams.page + 1 })}>Next Page</button>
   </div>
 );
+
+
+```
+### Guard
+
+When notLogin(in below) is true that mean it will not do anything but if it is string it will redirect to the string that we get in return.
+
+```js
+const notLogin: Accessor<string | true> = createMemo(() => {
+  if (JSON.parse(JSON.stringify(LoginService.get())).login != false) { return '/' }
+  return true;
+});
+const App: Component = () => {
+  return (
+    <>
+      <Router>
+        <Routes>
+          <Route path='/' component={Public_Layout}>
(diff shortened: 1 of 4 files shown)


puneetxp:
It seem i should stick to some my customization version of it. Thanks for your work.

PR #237, opened 2023-02-03
Update README.md

klener89:
Fixed the handling of the searchParams return type and convert it from String to Number in order to increment it. Default to 0 if parameter not defined yet. The handling of the empty page String in the setSearchParams feels cleaner than initializing the page to 0, as the parameter is mostly used after the first page.


Brian-Pob:
+1

I think this would be a good addition to the docs. Or at least indicate in the docs that a `parseInt` might be needed/ recommended.

PR #245, opened 2023-02-22
Add `exactActiveClass` to A component

emdede:
## The problem
With current design of the `A` component classes, there is one shortcoming:
An `A` component can be in one of three states: `inactive`, `active` and `exactActive`. But you cannot independently provide classes for each of the three since you have to choose between one of the two states `active` and `exactActive` by using the `end` property.

What this PR does is add the `exactActiveClass` prop which makes it so that you can provide classes for each of the three possible states. It also deprecates the `end` property.

## Considerations
Imo, treating `active` and `exactActive` as two distinct states is the wanted behavior because of a possible class pollution.

Simple example:
```
<A href="/main/nested" activeClass="text-white" exactActiveClass="text-red" />
```
With inclusive matching, this will generate an `a` element with class `text-white text-red`. It pollutes the space and also leads to class conflicts. A solution to make sure the `exactActive` classes are applied would be for the user to make them `important` but it's unintuitive and pollutes the class space even more.

Having this in mind, I think going for a stricter mutually exclusive distinction is the best way to go.
It does lead to duplication ( e.g. `activeClass="text-white text-lg" exactActiveClass="text-white text-xl"` ) but at least there is nothing unexpected going on in this scenario.

I'm open to suggestions and looking forward to feedback. Have a nice day

## Update
I've edited some of the original post to keep it a bit more concise. I went with a non-breaking implementation and one that is a bit more explicit than it was before. Mostly because of the added complexity of allowing `true` to just pass the `exactActiveClass` attribute. "3 distinct States" is opt-in.

It would now work like this (disregarding `inactive` class):
`<A href="/home" activeClass="text-white">Home</A>` will match both partially and exactly and class will be `text-white`

`<A href="/home" exactActiveClass>Home</A>` enables strict matching => class will either be `active` or `exactActive`

`<A href="/home" exactActiveClass="text-white">Home</A>` enables strict matching => class will either be `active` or `text-white`

This makes me think, wouldn't it be nice to have this same opt-in behavior for the other classes as well? Currently, there are `inactive` and `active` classes whether I need/want them or not. Would give a bit more control to users to have to opt-in first.

`<A href="/home" inactiveClass activeClass>Home</A>`

But I'll leave that for a different PR/time.

### DIFF
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -208,14 +208,14 @@ export interface AnchorProps extends Omit<JSX.AnchorHTMLAttributes<HTMLAnchorEle
   state?: unknown;
   inactiveClass?: string;
   activeClass?: string;
-  exactActiveClass?: string;
+  exactActiveClass?: true | string;
   /**
- * @deprecated end property deprecated in favor of 'exactActiveClass'
- */
+   * @deprecated end property deprecated in favor of 'exactActiveClass'
+   */
   end?: boolean;
 }
 export function A(props: AnchorProps) {
-  props = mergeProps({ inactiveClass: "inactive", activeClass: "active", exactActiveClass: 'exactActive' }, props);
+  props = mergeProps({ inactiveClass: "inactive", activeClass: "active" }, props);
   const [, rest] = splitProps(props, [
     "href",
     "state",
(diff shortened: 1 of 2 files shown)


emdede:
I realized that with my initial implementation, even though `end` was marked deprecated it didn't actually account for the new matching paradigm, so I removed it altogether for now.

src/components.tsx
```diff
@@ -239,11 +239,12 @@ export function A(props: AnchorProps) {
       state={JSON.stringify(props.state)}
       classList={{
         ...(props.class && { [props.class]: true }),
-        [props.inactiveClass!]: !isActive(),
-        [props.activeClass!]: isActive(),
+        [props.inactiveClass!]: !isActive()[0],
+        [props.activeClass!]: isActive()[0] && !isActive()[1],
```
emdede, line 243: Change to 
```
[props.activeClass!]: isActive()[0] && (!props.exactActiveClass || !isActive()[1]),
```
Would allow for loose matching if no `exactActiveClass` was provided.

bikeshedder:
How about deprecating the `end` property and adding both an `exactActiveClass` and `startActiveClass` instead?

That way it would not be a breaking change, but future versions of the library could remove the `end` property which is superseeded by the `exactActive` and `startActive` memos. `active` could just stay as is and be a combination of `exactActive | startActive`.

Alternative names for `startActive` could be `startsWithActive`, `prefixActive` or `parentActive`.

emdede:
Update:
It's non-breaking now. And I also went with a more explicit implementation. Mostly because of the added complexity of allowing `true` to just pass the `exactActiveClass` attribute.

It would now work like this:
`<A href="/home" activeClass="text-white">Home</A>` will match both partially and exactly and class will be `text-white`

`<A href="/home" exactActiveClass>Home</A>` enables strict matching => class will either be `active` or `exactActive`

`<A href="/home" exactActiveClass="text-white">Home</A>` enables strict matching => class will either be `active` or `text-white`

bikeshedder:
I like the names `exactMatch` and `partialMatch`. **<\/bikeshedding>**

ryansolid:
The reason this has not been merged is that it made me wonder if handling active class in the link makes sense at all anyway.  I do agree though the `end` property is less than ideal. The whole handling of what is active in the Anchor has led to all this permutation for what should be a simple link. So trying to think a bit wider outside of the box.

PR #246, opened 2023-02-25
Remove / trimming on the end of href

naturboy:
### The Problem
When using the `<A>` component the href value gets striped of any '/' at the end. 
Example:
`<A href={"/users/"}`>Users</A>
becomes 
`<a href="/users">Users</a> `
in the html. This behaviour is irritating for developers especially since the route could handle links with '/' at the end just fine. This PR is changing the behaviour of the normalise function to correct the behaviour.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,10 +2,10 @@ import { createMemo, getOwner, runWithOwner } from "solid-js";
 import type { MatchFilter, MatchFilters, Params, PathMatch, Route, SetParams } from "./types";
 
 const hasSchemeRegex = /^(?:[a-z0-9]+:)?\/\//i;
-const trimPathRegex = /^\/+|\/+$/g;
+const trimPathRegex = /^\/+|(\/)\/+$/g;
 
 export function normalizePath(path: string, omitSlash: boolean = false) {
-  const s = path.replace(trimPathRegex, "");
+  const s = path.replace(trimPathRegex, "$1");
   return s ? (omitSlash || /^[?#]/.test(s) ? s : "/" + s) : "";
 }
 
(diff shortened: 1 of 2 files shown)


Brendan-csel:
I think the challenge here is `/users/` and `/users` ultimately need to match the same route.  The one that has caused more issues in the past is the root url `http://mysite.com` versus `http://mysite.com/`.  Or if there is a `baseUrl` in configuration.  Also, some servers append the trailing slash on initial page load and others don't - and of course folks can link to either from outside.

The `normalizePath` function is not only used for `<A>`.  It is also used to normalize the route when matching - so it really needs to result in the same output whether there is a trailing slash in the input or not.  It is also used when joining path parts where it needs to strip the trailing slash to avoid double slashes.

If optionally keeping the trailing slash in the `<A>` href is really a concern then it probably needs its own regex.  It would be good to keep the code located close to the existing `normalizePath` function though as they need to (almost) "match".

naturboy:
I understand the problem that '/users/' and '/users' must finally point to the same route. However, I think the way the routes look inside the application should be chosen by the developer. It leads to a bad experience if links don't look like the developer described them in his code. In addition, in some frameworks the router makes a distinction between '/users/' and '/users'. 
In my tests the router can already handle '/users/ and '/users' without problems even with the changes to the 'normalizePath' function. Also I could not find any situation where double slashes occur. Even when merging paths it should not happen, because all double slashes are always removed at the beginning and only one remains at the end. This can also be seen in utils.spec.ts in line 243. 
So to me it seems like the changes should be fine. Could you construct a case where my changes break the router?

DominicTobias:
> I think the challenge here is `/users/` and `/users` ultimately need to match the same route. The one that has caused more issues in the past is the root url `http://mysite.com` versus `http://mysite.com/`.

Sorry to hijack but in that case do you know how I can get `activeClass` to match both `/path` and `/path/` in combination with `end` (since I don't want to match `/path/something`?

PR #279, opened 2023-07-10
feat: intent based route preloading

marbemac:
This PR adds `preload`  (similar to `remix`) and `preloadDelay` (similar to `tanstack-router`) options to the `Router` and `Link` components.

I'm slowly making my way over from the world of React, so the particular implementation I chose might not be ideal (more on this in comments). I tried to minimize the amount of change, and it seemed a little awkward to get access to the relevant route data and asset preload functions in the right spots given how things are currently organized.

- [x] No breaking changes
- [ ] No tests added - figured I'd wait to get feedback on whether this PR might be accepted or not before spending time here
- [ ] Ditto re docs - can update the readme if this is something that might get merged

**Usage**

```tsx
// Set global defaults via Router
<Router preload="intent" preloadDelay={500} />

// Override global default on a per Link basis
<Link preload="intent" preloadDelay={500} />
```

If the target route has a data function AND/OR if it has a lazy loaded component, the data func and lazy assets will be prefetched.

### DIFF
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -1,7 +1,16 @@
 /*@refresh skip*/
 
 import type { Component, JSX } from "solid-js";
-import { children, createMemo, createRoot, mergeProps, on, Show, splitProps } from "solid-js";
+import {
+  children,
+  createMemo,
+  createRoot,
+  mergeProps,
+  on,
+  onCleanup,
+  Show,
+  splitProps
+} from "solid-js";
 import { isServer } from "solid-js/web";
 import { pathIntegration, staticIntegration } from "./integration";
 import {
@@ -19,6 +28,7 @@ import {
   useRouter
 } from "./routing";
 import type {
+  LinkPreloadOpts,
   Location,
   LocationChangeSignal,
   MatchFilters,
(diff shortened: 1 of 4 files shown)


ryansolid:
Thank you. This is interesting. Sorry it took me so long to get back here. I am not sure if you saw the work Ryan Turnquist did on https://github.com/solidjs/solid-router/compare/main...next

He was the original co-author of this library and started on a preload approach but I was never sure if this is how we wanted to do it. The problem with re-running the data functions is it does nothing unless we cache the results.. we had decided we'd leave that to the library, but we never had a blessed solution so the feature didn't really work well.

We talked about keeping the results of the data function around and then passing them into the actual render as available as that wouldn't waste work, but they would be live queries at that point so managing their lifecycle might be hard. I'm gathering you are doing the latter. Can you describe a bit more about the behavior? Am I correct in my understanding? LIke if something is preloaded at what point to do we throw it away?

marbemac:
No worries! Ah no I hadn't seen Ryan's work.

Yeah... imho data caching prob should not be a concern of solid-router. However, this opt-in preload behavior gives end users the option to combine solid-router with other libs that handle data caching - for example this functionality pairs very nicely with solid-query.

ryansolid:
Happy to say this will be resolved by #312

PR #291, opened 2023-08-23
💥 !Define `searchParamIntegration`! 💥

devinrhode2:
Context: I am building a SolidStart app which will be embedded into other non-Solid-based apps by simply injecting the html+scripts of the SolidStart app at runtime. Once an initial URL is constructed for the SolidStart app, a component similar to [this](https://github.com/justinfagnani/html-include-element/pull/24) is used to inject the scripts + html. The html+scripts come from my SSR'd SolidStart app.

Context2: To make this SolidStart "widget" or "micro-frontend" truly isomorphic, it needs to add some state into the url. (Cookies could be another option). This of course allows the server to SSR the right page.

This PR creates a new way for Solid apps to handle routing, where the pathname and search params for a solid app are encoded into a single search param. This allows Solid apps to be embedded into larger apps without adversely affecting their url.

Host url: `/hPath1/hPath2?hParam1=hParam1Value#hHash`
Solid app's url: `const solidUrl = '/sPath1/sPath2?sParam1=sParam1Value#sHash'`

Host app's url, with Solid app embedded inside it:
1. Set query param: `searchParamIntegration('CARNIATOMON')`
2. Encode it: `/hPath1/hPath2?hParam1=hParam1Value&CARNIATOMON=${encodeURIComponent(solidUrl)}#hHash`
3. Fully encoded: `/hPath1/hPath2?hParam1=hParam1Value&CARNIATOMON=%2FsPath1%2FsPath2%3FsParam1%3DsParam1Value%23sHash#hHash`

### DIFF
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -178,15 +178,15 @@ export function hashIntegration() {
       go: delta => window.history.go(delta),
       renderPath: path => `#${path}`,
       parsePath: str => {
+        // Get everything after the `#` (by dropping everything before the `#`)
         const to = str.replace(/^.*?#/, "");
-        // Hash-only hrefs like `#foo` from plain anchors will come in as `/#foo` whereas a link to
-        // `/foo` will be `/#/foo`. Check if the to starts with a `/` and if not append it as a hash
-        // to the current path so we can handle these in-page anchors correctly.
         if (!to.startsWith("/")) {
+          // We got an in-page heading link.
+          // Append it to the current path to maintain correct browser behavior.
           const [, path = "/"] = window.location.hash.split("#", 2);
           return `${path}#${to}`;
         }
-        return to;
+        return to; // Normal Solidjs <A> link
       }
     }
   );
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -160,6 +160,68 @@ export function pathIntegration() {
   );
 }
 
+/**
+ * If your Solidjs "app" is really just a "widget" or "micro-frontend" that will be embedded into a larger app,
+ * you can use this router integration to embed your entire url into a *single* search parameter.
+ *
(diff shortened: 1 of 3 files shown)


src/integration.ts
```diff
@@ -149,6 +160,69 @@ export function pathIntegration() {
...
+    ({ value, replace, state }) => {
+      if (replace) {
+        window.history.replaceState(state, "", value);
+      } else {
+        window.history.pushState(state, "", value);
+      }
+      // Because the above `pushState/replaceState` call should never change `window.location.hash`,
+      // this behavior from `pathIntegration` doesn't make sense in the context of `searchParamIntegration`:
+      //   scrollToHash(window.location.hash.slice(1), scroll);
+      // Perhaps when a widget loads, it should run the `el.scrollIntoView` function itself.
```
devinrhode2, line 205: <sub>Although url hashes in the Solid widget url are not disallowed, I wasn't sure how the hash-scrolling behavior could be carried over into this new context. Let's say we have a Solid app we want to embed into another app. The scripts get injected. This "widget" is handling the entire screen between the header and footer. Therefore, it may have `<h1>` links inside the page. In this case, perhaps `searchParamIntegration` should get some new functionality. If a Solid.js-based micro-frontend can agree to a contract with the parent/host app that the parent/host app will not be using url hashes when the Solid.js-based micro-frontend is active, then `searchParamIntegration` could take in an extra parameter to indicate that url hashes should not be encoded into the search param, but instead independently added/removed to url. Given this contract and some new parameters passed to `searchParamIntegration`, `searchParamIntegration` could then mostly copy the scroll behavior of `pathIntegration`.</sub>

devinrhode2:
I guess I'm pioneering... "isomorphic micro-frontends"... 🤣

PR #312, opened 2023-12-06
Huge Router Refactor + New Data Paradigm + Preloading

ryansolid:
This PR implements the RFC I posted a few weeks back: https://github.com/solidjs/solid-router/discussions/306

See the updated Readme: https://github.com/solidjs/solid-router/blob/next/README.md

I believe I've implemented the RFC plus some learnings and incorporated the feedback.

Just doing some final testing/finishing touches.


README.md
```diff
@@ -439,66 +323,248 @@ function PageWrapper() {
...
+
+getUser.key // returns "users"
+getUser.keyFor(id) // returns "users[5]"
 ```
 
-## Memory Mode Router
+You can revalidate the cache using the `revalidate` method or you can set `revalidate` keys on your response from your actions. If you pass the whole key it will invalidate all the entries for the cache (ie "users" in the example above). You can also invalidate a single entry by using `keyFor`.
 
-You can also use memory mode router for testing purpose.
+`cache`` can be defined anywhere and then used inside your components with:
```
fezproof, line 383: Minor typo with extra tick
```suggestion
`cache` can be defined anywhere and then used inside your components with:
```

README.md
```diff
@@ -533,32 +599,100 @@ const routes = [
...
+<Route path="/redirect" component={() => <Navigate href={getPath} />} />;
+```
+
+### `<Route>`
+
+The Component for defining Routes:
+
+| prop | type | description |
+|-|-|-|
+|TODO
```
gbj, line 694: missing docs for `<Route>` props

gbj:
One higher-level thing I'm curious about: This shifts things like `<Outlet>` away from using context, allowing you to render nested routes on the server without rendering the whole page etc., but `<A>` still relies on context for its resolved path/active route matching, `useNavigate()` and other hooks rely on context, etc. I'm curious how these work or don't with islands: i.e., if you `useNavigate()` in an island does it throw, or is the plan to have an island router context provider in the future, or something else?

PR #314, opened 2023-12-07
Add RouteDefinition to BaseRouterProps

iczero:
Allows use as
```jsx
let routes = [{ path: '/', component: Whatnot }];
return <Routes>{routes}</Routes>;
```
without TypeScript complaining.

I am not sure if this is the right solution.


ryansolid:
Seems reasonable. Thank you.

Issue #277, opened 2023-06-27
useBeforeLeave not firing on history.back()


KamehamehaNudel:
### Describe the bug

The "useBeforeLeave" hook seems to only be firing when navigating using links.
When using the browsers back/forth buttons it does not appear to execute.

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-ejhtft?file=src%2FApp.jsx

### Steps to Reproduce the Bug or Issue

1. Go to the "test" route using the nav link.
2. Go back to the "home" routeusing the nav link.
3. An alert "here" will show up
4. Go back to the "test" route using the nav link.
5. Click the "back" button (fires history.back())
6. The route changes correctly, but the alert does not show up 

### Expected behavior

As a user i expeted the "useBeforeLeave" function to also fire when using the "back" button (either progromatically or with the actual browser buttons in a 'real' environment).
I am in both cases leaving the current route so I don't see why it shouln't execute.

### Platform

- OS: Windows
- Browser: Firefox
- Version: 114.0.2 (64-Bit)

Brendan-csel:
The [PR that added useBeforeLeave](https://github.com/solidjs/solid-router/pull/186) only supports router-initiated blocking.  Not browser-initiated blocking such as back/forward buttons.

I do have a [fork of this router](https://github.com/cselnz/solid-router/tree/publish) were I had a go at adding those other blockers.  We have been using that in production and it seems to be working fine - but hasn't had broader testing.  Also we us the Hash Integration so "normal" integration hasn't been tested.  I think [this is the commit](https://github.com/solidjs/solid-router/commit/51b050403462225a2997d10c87d1a97073027d57) that adds that ability.

KamehamehaNudel:
Aye i See. Thanks for the quick answer, I'll see if i can get that running as a custom integration or patch.
Will that also make it to main version (eventually) ?
Or is it 'intentional' afterall?

Brendan-csel:
I just happened update that branch of mine recently so if you want to quickly test with it you can try install `csel-solid-router` from npm and find/replace to update any files that import the router.

> Will that also make it to main version (eventually) ? Or is it 'intentional' afterall?

That'll be up to @ryansolid.  It is quite a bit of code - although it "works for me".

CreativeTechGuy:
I just want to chime in that I found this issue because this is something that I was very surprised was missing. Hopefully it can get added!

ryansolid:
[...]

PR #320, opened 2023-12-10
Extend useBeforeLeave for browser initiated events (eg back/forward)

Brendan-csel:
This PR extends `useBeforeLeave` blocking to include router initiated navigations such as back/forward.

See issue https://github.com/solidjs/solid-router/issues/277

We've been using this approach in production for over year now - albeit with the `0.8` router generation and hash integration.  Today I rebased this onto `0.10.1` and did some basic testing.

One aspect that might be contentious is using `_depth` on history state to detect navigation type.

Even if not merged, hopefully this PR offers some guidance for others needing this feature.

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -1,4 +1,5 @@
-import { BeforeLeaveLifecycle, BeforeLeaveListener, NavigateOptions } from "./types";
+import { isServer } from "solid-js/web";
+import { BeforeLeaveLifecycle, BeforeLeaveListener, LocationChange, NavigateOptions } from "./types";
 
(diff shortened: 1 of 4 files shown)


ryansolid:
This is definitely a feature that I feel is missing from the router. The contentious part you are talking about is backing out of the app right? Especially after browser refresh as you won't have the history anywhere else? I've used something like depth in the past to solve this so it seems reasonable to me.

What's the concern?

Brendan-csel:
> What's the concern?

Really just this pollutes `window.history.state` with a `_depth` property to enable this feature.

PR #326, opened 2023-12-17
fix action types

brenelz:
I was getting a typescript error when returning a value from an action and using the `with` function

image

I think this fixes it. I also think you'd want to update to a newer solid so the action types work better.


ryansolid:
Yep. Thank you.

PR #331, opened 2023-12-20
Improve `Router` component's docs

femincan:
- The types for the `children` prop of the `Router` component have been updated to align with the PR: #314.
- A New documentation section was added in the `Config Based Routing` section to show the usage of a single router definition object.


ryansolid:
Thanks

PR #333, opened 2023-12-21
fix: make search params return type partial

### LINKED ISSUE #329: [Bug]: `useSearchParams<{foo: string;}>` generic should pass through `Partial<>`
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Calling `useSearchParams<{foo: string}>` should make the returned `searchParams` type `{ foo?: string }`. This is because we never know what kind of query params somebody will actually send.

### Expected behavior 🤔

See previous

### Steps to reproduce 🕹

assert that `ReturnType<typeof useSearchParams<{ foo: 'foo';}>>[0]` has type `string | undefined`

### Context 🔦

_No response_

### Your environment 🌎

_No response_

femincan:
The `useSearchParams` hook's return type now wrapped with `Partial`. Fixes #329.

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -90,7 +90,7 @@ export const useMatch = <S extends string>(path: () => S, matchFilters?: MatchFi
 export const useParams = <T extends Params>() => useRoute().params as T;
 
 export const useSearchParams = <T extends Params>(): [
-  T,
+  Partial<T>,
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

PR #336, opened 2023-12-24
docs: change root element ids from app to root

femincan:
The root element ids in the examples changed from `app` to `root`. Because when we create a new solid-js app, vite uses `root` as the default.


ryansolid:
Hmm.. and SolidStart uses app and will replace those templates eventually. Its a good thought, but I'm going to leave it as is for now.

PR #339, opened 2023-12-27
feat: expose params on event for use in middlware

brenelz:
This would allow you access to url params in the middeware.  Saw someone ask for it and figured this out.

```typescript
export default createMiddleware({
    onBeforeResponse: [
        event => {
            console.log(event.context.params?.id);
        },
    ]
});
```


ryansolid:
Interesting. Although it would only be available after rendering. That being said .context is an H3 thing. We need to thing about where this goes.. standardizing it etc.

brenelz:
Sounds good I can close this for now. It was just an idea

Issue #344, opened 2024-01-02
Typing fails to detect data prop type


staadecker:
### Describe the bug

The current type setup requires a lot of extra code. Specifically in the following example I need to
1.  cast the component using `as Component` and

2. I need to check that `props.data` is not undefined even though it cannot be (afaik).

```javascript
// App.tsx
const fetchData = async () => {
  await new Promise((resolve) => setTimeout(resolve, 5000));
  return 'Fetched the data!';
};

const loadData = () => {
  const [res] = createResource(fetchData);
  return res;
};

const App: Component = () => {
  return (
    <Router>
      <Route path="/" component={HomePage as Component} load={loadData} /> // 👈 Casting shouldn't be needed
    </Router>
  );
};
```

```javascript
// HomePage.tsx
export const HomePage = (props: RouteSectionProps<Resource<string>>) => {
  if (props.data === undefined) { // 👈 This shouldn't be needed...
    throw Error('Something unexected happened!');
  }

  return <Suspense fallback={<p>Loading...</p>}>{props.data()}</Suspense>;
};
```

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-ssnat2?file=src%2FApp.tsx

### Steps to Reproduce the Bug or Issue

1. Remove the `as Component` and observe how typescript throws an error
2. Or remove the `props.data === undefined` if statement and observe how typescript throws an erro

### Expected behavior

Neither should be needed

### Platform

- OS: Windows
- Browser: Firefox
- Version: 0.10.5

staadecker:
PR #345 fixes issue 1. here but not number 2.

Brendan-csel:
`props.data` will be `undefined` unless your route's `load` function happens to return something.

staadecker:
> `props.data` will be `undefined` unless your route's `load` function happens to return something.

aha gotcha. I imagine it would be tricky (or completely impossible) to encode that logic in the typescript type system.

PR #345, opened 2024-01-02
Partial fix for #344

staadecker:
I believe this is a partial fix to #344 . Now a `Route` can have props of type `RouteProps<S, T>` rather than just `RouteProps<S, undefined>`


ryansolid:
Thanks. yeah added the data prop last min and didn't propagate it through everywhere.

Issue #349, opened 2024-01-04
Form action set as insecure by google chrome.


AlbertSabate:
### Describe the bug

When accessing a website with https, the forms action="" that are not https:// are set as insecure by Google Chrome [Google Chrome Docs](https://blog.chromium.org/2020/08/protecting-google-chrome-users-from.html).

The new form action on @solidjs/router renders the markup like this:
```{html}
<form method="post" action="action:signUpForm">...</form>
```

Google Chrome sets it as insecure, as shown in the image below.
<img width="787" alt="Screenshot 2024-01-04 at 11 04 24 PM" src="https://github.com/solidjs/solid-router/assets/2384218/ed86c3a1-628f-4e2c-8df7-bac683342d1d">

### Your Example Website or App

https://start.solidjs.com/core-concepts/actions#using-forms-to-submit-data

### Steps to Reproduce the Bug or Issue

Create a form using the example on the solidjs/start a website.
Open the site using https.
Click on any input, and the message will show.

### Expected behavior

The form should not be set as insecure as it could generate distrust from the users.

### Platform

Google Chrome.

ryansolid:
Alright 0.10.7 is released

PR #350, opened 2024-01-04
Form action set as insecure by google chrome.

AlbertSabate:
Fixes issue #349

When accessing a website with https, the forms action="" that are not https:// are set as insecure by Google Chrome [Google Chrome Docs](https://blog.chromium.org/2020/08/protecting-google-chrome-users-from.html).

The new form action on @solidjs/router renders the markup like this:

`<form method="post" action="action:signUpForm">...</form>`

This PR changes the action to:

`<form method="post" action="https://action/signUpForm">...</form>`

Preventing Google Chrome from seeing the action as unsafe.

### DIFF
--- a/src/data/action.ts
+++ b/src/data/action.ts
@@ -97,8 +97,8 @@ export function action<T extends Array<any>, U = void>(
 
   const url: string =
     (fn as any).url ||
-    (name && `https://action:${name}`) ||
-    (!isServer ? `https://action:${hashString(fn.toString())}` : "");
+    (name && `https://action/${name}`) ||
+    (!isServer ? `https://action/${hashString(fn.toString())}` : "");
   return toAction(mutate, url);
 }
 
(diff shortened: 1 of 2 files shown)


ryansolid:
Hmm.. only thing I'm wondering is if we are lying here if the protocol for the site isn't https..  Then again this whole thing is a lie in a sense as we never actually submit the form. I'm just wondering if we just used `/action` without the https would this still work on an https site.

I guess the biggest issue there would be eliminating the base URL. Or I guess using the attribute value instead of the property value. 

I can merge the PR and play with it. I guess there is no easy way to test this locally.

AlbertSabate:
Hello @ryansolid,

I can make that change, no worries; I got the setup done already. Because we are attaching an event doing preventDefault, it won't conflict with a real route starting for /action (I hope :D)

Let me amend the change and check if it works properly.

PR #358, opened 2024-01-21
Use `.getAttribute('action')` instead of `.action`

### LINKED ISSUE #357: actionRef.startsWith is not a function
### Describe the bug

In `handleFormSubmit` it appears that somehow `evt.target.action` is an `HTMLButtonElement`, rather than a string.

https://gitlab.com/letschurch/lets.church/-/tree/web-solid-next/apps/web-solid-next?ref_type=heads

Using the `action` function as your `action` with `method` set to `'post'` can somehow cause `evt.target.action` to be a button rather than a string, and the call to `startsWith` fails.

### Expected behavior

Using the `action` function as `action` with `method` set to `post` should always work.

### Screenshots or Videos

image

image

image

### Platform

- OS: macOS
- Browser: Firefox
- Version: 121

### Additional context

I don't really know how this is happening. I would expect `.action` on an `HTMLFormElement` to always be a string (or undefined?) but it's a button. In this case, as you can see in the screenshot above, using `.getAttribute` returns the right value.

I have other forms that are working just fine, I don't know what's going on with this particular form.

knpwrs:
Fixes #357

### DIFF
--- a/src/data/events.ts
+++ b/src/data/events.ts
@@ -99,7 +99,7 @@ export function setupNativeEvents(preload = true, explicitLinks = false, actionB
       let actionRef =
         evt.submitter && evt.submitter.hasAttribute("formaction")
           ? (evt.submitter as HTMLButtonElement | HTMLInputElement).formAction
-          : (evt.target as any).action;
+          : (evt.target as any).getAttribute('action');
       if (!actionRef) return;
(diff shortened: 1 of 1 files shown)


ryansolid:
Ok I see what I need to do. Thank you

PR #392, opened 2024-03-22
Fix types for cache functions with all optional arguments

Brendonovich:
`cache` functions with all optional arguments are currently casted to `(...args: never[]) => T` as `[]` apparently extends tuples with all optional values. This fixes that by doing the `[] extends` comparison against the argument list with optionality removed. This should be fine since the only important part of that check is whether there are >0 arguments, we don't really care about the types.


ryansolid:
I don't know better but I trust you.

Issue #396, opened 2024-03-28
`useLocation().state` does not get updated on history back


r-Larch:
### Describe the bug

When navigating to a route with `state` object the route correctly sees the state but when navigating back the location.state will not update.

### Your Example Website or App

https://stackblitz.com/edit/solidjs-templates-o9qpzf?file=src%2Findex.tsx

### Steps to Reproduce the Bug or Issue

When navigating to a route with `state` object the route correctly sees the state but when navigating back the location.state will not update.

```js
// having a component on route '/' with:
const location = useLocation();
createEffect(() => {
  console.log(location.state)
})

// doing with a button click:
// 1. navigate to:
navigate('/', { state: { foo: 'bar' } }); //   console: { foo: 'bar' }
// 2. navigate to:
navigate('/', { state: { foo: 'baz' } }); //   console: { foo: 'baz' }
// 3. navigate back:
history.back(); //   console: <no change> !! expected { foo: 'bar' }
```

Demo Reproduction: https://stackblitz.com/edit/solidjs-templates-o9qpzf?file=src%2Findex.tsx

### Expected behavior

`useLocation().state` should reflect `history.state` and should reactively update.

### Platform

- OS: Windows
- Browser: Chrome
- Version: "^0.13.1"

r-Larch:
I think changing the `equals` check in `createRouter.ts` line: 35 could be a fix.
The equals function only checks the path and does not check if state has changed.

https://github.com/solidjs/solid-router/blob/e773947b85ac78281816e86621a2cdb6735ae95a/src/routers/createRouter.ts#L25-L41

PR #405, opened 2024-04-15
fix: state updates detection on history back/forward

### LINKED ISSUE #275: Undetected location state changes while using browser back/forward buttons
### Describe the bug

Navigation to the same route path with different state is detected using `navigate(path, { state: ...})` directly — however when using browser back/forward buttons, the state changes are not detected until the page is refreshed.

https://stackblitz.com/edit/solid-router-back-forward-undetected-state-change

1. Navigate to the StackBlitz repro link 
2. Wait for the installation to finish and the dev server to start
3. **Important** Open the preview in a new tab `Open in New Tab`
4. Initial state should be `null`
5. Click on `Go to about with active true` button
6. You'll see the page updating to `/about` and the state is `{"active": true}`
7. Click on the browser `back` button
8. You'll see the home page `/` and the state is `null`
9. Click on `Go to home with active false` button
10. You'll see the home page `/` and the state is `{"active":false}`
11. Click on the browser `back` button
12. You'll still see the home page `/` and the state is `{"active":false}` **the unexpected behavior**
13. Refresh the page
14. You'll see the home page `/` and the state is `null` **expected at 12. without refresh**

### Expected behavior

Same steps as above in the reproduction, starting from 11:

11. Click on the browser `back` button
12. See the home page `/` and the state is `null` without refreshing the page

Same with the browser `forward` button

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Brave - Chromium based
- Chromium Version: `114.0.5735.133`
- Brave Version: `1.52.126`

### Additional context

_No response_

oedotme:
This PR adds an event listener to `popstate` change event to update the router state with the history back/forward events between two routes with the same paths but with different state objects.

Currently the changes are only detected by refreshing the page.

Closes #275 and #396

### DIFF
--- a/.changeset/lucky-dingos-happen.md
+++ b/.changeset/lucky-dingos-happen.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Fix state updates detection on history back/forward
--- a/src/routers/createRouter.ts
+++ b/src/routers/createRouter.ts
@@ -1,7 +1,6 @@
 import { Signal, createSignal, onCleanup } from "solid-js";
 import type { LocationChange, RouterContext, RouterUtils } from "../types.ts";
 import { createRouterComponent } from "./components.jsx";
-import { equalObjects } from "../utils.js";
 
 function intercept<T>(
   [value, setValue]: [() => T, (v: T) => void],
@@ -34,7 +33,7 @@ export function createRouter(config: {
   const wrap = (value: string | LocationChange) => (typeof value === "string" ? { value } : value);
   const signal = intercept<LocationChange>(
     createSignal(wrap(config.get()), {
-      equals: (a, b) => a.value === b.value && equalObjects(a.state, b.state)
+      equals: (a, b) => a.value === b.value && a.state === b.state
     }),
     undefined,
     next => {
--- a/src/routers/createRouter.ts
+++ b/src/routers/createRouter.ts
@@ -1,6 +1,7 @@
 import { Signal, createSignal, onCleanup } from "solid-js";
 import type { LocationChange, RouterContext, RouterUtils } from "../types.ts";
 import { createRouterComponent } from "./components.jsx";
+import { equalObjects } from "../utils.js";
 
 function intercept<T>(
   [value, setValue]: [() => T, (v: T) => void],
@@ -32,7 +33,9 @@ export function createRouter(config: {
   let ignore = false;
   const wrap = (value: string | LocationChange) => (typeof value === "string" ? { value } : value);
   const signal = intercept<LocationChange>(
-    createSignal(wrap(config.get()), { equals: false }),
+    createSignal(wrap(config.get()), {
+      equals: (a, b) => a.value === b.value && equalObjects(a.state, b.state)
+    }),
     undefined,
     next => {
       !ignore && config.set(next);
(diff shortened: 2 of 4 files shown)


ryansolid:
I see what you are trying to do but could this already be handled by the current popstate event? I'm gathering this is related to that we don't have access to the state representation where we wire the events.

I'm just looking at:
https://github.com/solidjs/solid-router/blob/3cd2fef67fe0b4d8bb3178923be32d541d56ec79/src/routers/Router.ts#L29-L38

And more specifically:
https://github.com/solidjs/solid-router/blob/3cd2fef67fe0b4d8bb3178923be32d541d56ec79/src/lifecycle.ts#L64-L77

And thinking there should be some way to set the source state properly and ensure that the notifications are working properly. Like the bug seems most likely with our way of telling things have changed rather than sideloading in a second event.

oedotme:
@ryansolid Thanks for your response, and yeah that makes sense.

From what I found so far, the source `signal` doesn't trigger updates when the state changes while the `path` value is the same, because of the following `equals` condition:

https://github.com/solidjs/solid-router/blob/3cd2fef67fe0b4d8bb3178923be32d541d56ec79/src/routers/createRouter.ts#L35

To allow the state change detection, we can set `equals` to:
- `false`
- or `(a, b) => a.value === b.value && JSON.stringify(a.value) === JSON.stringify(b.value)`
- or even `JSON.stringify(a) === JSON.stringify(b)`

I went with `false` to avoid `JSON.stringify` potential performance hits, but please let me know what you think:

```diff
-       createSignal(wrap(config.get()), { equals: (a, b) => a.value === b.value }),
+       createSignal(wrap(config.get()), { equals: false }),
```

<br>
<br>

The other related logic that skips updates if the `path` hasn't changed:

https://github.com/solidjs/solid-router/blob/3cd2fef67fe0b4d8bb3178923be32d541d56ec79/src/routing.ts#L333-L343

I was unsure about resetting error-boundaries and submission here as well, so I just added an `else` statement in order to handle the state-only changes:

```diff
+      } else {
+        start(() => {
+          intent = "native";
+          setState(state);
+        }).then(() => {
+          intent = undefined;
+        });
      }
```

Those two changes solved the state-only detection issue. Please let me know what you think, thanks!

ryansolid:
I see.. hmm.. yeah that equality check I guess needs to fail here but I do wonder if this can cause other issues. Shortcuts when trying to navigate to the same route. Probably nothing to sever.

We should reset errorboundaries and submissions in this case, because it is a navigation.

oedotme:
@ryansolid I re-added the `equals` check again just in case, to avoid any potential unexpected behaviors.

I also added an object equality comparison utility based on [fast-deep-equal](https://github.com/epoberezkin/fast-deep-equal) to compare the `state`, which addresses the performance concern I initially had with `JSON.stringify`.

Please let me know what you think. Thanks!

ryansolid:
I'm going to play around with this and see if the equal check is important. I'm as concerned with adding more code as I am with JSON.stringify performance. So if we can avoid doing much here. My gut here is if the state reference is different that will be sufficient. It might be aggressive but it is probably fine.

oedotme:
@ryansolid Checking the state reference makes total sense in this case.

> Could it not just be?
> ```ts
> equals: (a, b) => a.value === b.value && a.state === b.state }
> ```

Yes, I've update the PR accordingly. Thanks again.

ryansolid:
Looks great thanks

PR #409, opened 2024-04-24
Form component

SarguelUnda:
# Summary

This PR does 2 things:
- better method detection for action. If the formmethod attribute is present, use it instead of the form element.
- add a Form component that could be compared to the A element i.e. does spa routing instead of native browser behaviors

## Consideration

To achieve this we use the same trick as action which is to globally register the form element in a Set. We register it by ref but we could consider to use solidjs createUniqueId if we prefere to use string instead of dom node.

## Example usage

Example app: https://github.com/SarguelUnda/solidrouter-pr-form

```tsx
import { Form, action } from '@solidjs/router';

const myAction = action(async (data: FormData) => {
  console.log("ACTION")
});


const MyForm = () => {
  return <Form action={myAction} method="post">
    <input name="foo" />
    <input name="bar" />
    <button type="submit">POST ME</button>
    <button name="getsubmit" type="submit" formMethod='get' formAction="newr?azeizaei=1" >GET ME</button>
  </Form>
}
```

### DIFF
--- a/src/data/action.ts
+++ b/src/data/action.ts
@@ -18,6 +18,8 @@ export type Action<T extends Array<any>, U> = (T extends [FormData] | []
 
 export const actions = /* #__PURE__ */ new Map<string, Action<any, any>>();
 
+export const routableForms = /* #__PURE__ */ new Set<HTMLFormElement>();
+
 export function useSubmissions<T extends Array<any>, U>(
   fn: Action<T, U>,
   filter?: (arg: T) => boolean
--- a/src/data/components.tsx
+++ b/src/data/components.tsx
@@ -1,0 +1,18 @@
+/*@refresh skip*/
+import type { JSX } from "solid-js";
+import { onCleanup, onMount, splitProps } from "solid-js";
+import { routableForms } from "./action.js";
+
+export interface FormProps extends JSX.FormHTMLAttributes<HTMLFormElement> {}
+export function Form(props: FormProps) {
+  const [, rest] = splitProps(props, ["ref"]);
+  onMount(() => {
+    routableForms.add(formRef)
+  })
+  onCleanup(() => routableForms.delete(formRef))
+  let formRef: HTMLFormElement;
+  return <form {...rest} ref={(el) => {
+    props.ref && (props.ref as Function)(el);
+    formRef = el
+  }} />
+}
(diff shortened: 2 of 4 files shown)


SarguelUnda:
Other points that might warrant discussion around this PR. The answer I'm giving are my own as a starter of conversation and any feedback on them is welcome

**question:** Why using a component and not extending the action api ?
 **answer:** We could extend the action api to register the form as a routable get form if instead of getting an async function it get a relative url as a string for instance. I didn't go toward this design mainly to not change an existing api, because I was not sure of all the implications. It also fell less natural to me on the userside of things but this is highly debatable. I can rewrite this pr if the other design is preferable

```tsx
import { Form, action } from '@solidjs/router';

const myAction = action(async (data: FormData) => {
  console.log("ACTION")
});


const MyForm = () => {
  return <Form action={myAction} method="post">
    <input name="foo" />
    <input name="bar" />
    <button type="submit">POST ME</button>
    <button name="getsubmit" type="submit" formMethod='get' formAction="newroute" >GET ME</button>
  </Form>
}

// vs 

const MyForm = () => {
  return <form action={myAction} method="post">
    <input name="foo" />
    <input name="bar" />
    <button type="submit">POST ME</button>
    <button name="getsubmit" type="submit" formMethod='get' formAction={action("newroute")} >GET ME</button>
  </form>
}
```

**question:**  Why the component is inside its own file and not with the other components ?
 **answer:** Data api as I understand is quite optional so I didn't want to write it within the global scope of the package.

SarguelUnda:
Hello,

Any advice how to bring this proposition forward ?
- Is this component better on the App side than on the library side ?
- Does this design conflict with futures direction of the solid router / data api ?
- Is it useful if I also propose the alternative without a component but by extending the action api to compare solutions ?
- Does some consideration about ssr / solid start escape me ? I only use solid in spa context.

ryansolid:
I admit I missed the point of suggestion. Hard to tell what the value is. There was a discussion about this issue in Discord but I can't seem to find it now. My understanding if I remember was that this would help do client side navigation similar to a Anchor tag by supporting GET method. But then I don't quite understand how actions play into it.

Basically is the target the client routed path or the action.. if it is the action then we can support GET encoding without a capital Form Component in the future. If it is the client routed path then actions don't fit exactly. I guess in a sense the client Router views itself as the GET for all routes it defines so only POST would make sense.

As I said I might need a much clearer explanation step by step about how this would behave.

PR #411, opened 2024-04-25
doc(cache): add missing async keyword

marvin-j97:
The await could be removed in favour of:

```ts
return fetch(...).then(res => res.json());
```


atila-crabnebula (approved):

atilafassina (approved):
this PR is so good I ended up approving it twice. 😅

PR #414, opened 2024-05-06
Fix no-revalidate from server action

LexSwed:
When `action` returns
```tsx
  return json(data, { revalidate: [] })
```

Empty revalidation keys array is serialised into an empty string, so `response.headers.get("X-Revalidate") === ''`.
This makes

https://github.com/solidjs/solid-router/blob/57847624fb673003f6f0d8261ca7352ff00e7480/src/data/action.ts#L152

to be an array of `['']` ([Playground](https://playground.solidjs.com/anonymous/7658a2ef-8611-4916-a007-715038b008a4)), causing all cache keys to be revalidated.

This PR adds empty string check, defaulting to an empty array, as empty array seem to correctly stop revalidation from happening.

Very subtle behaviour that should be tested, but I couldn't find a setup for `actions` testing.

### DIFF
--- a/.changeset/thin-baboons-kiss.md
+++ b/.changeset/thin-baboons-kiss.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Fix no-revalidate from server action
--- a/src/data/action.ts
+++ b/src/data/action.ts
@@ -149,7 +149,9 @@ async function handleResponse(response: unknown, error: boolean | undefined, nav
   let flightKeys: string[] | undefined;
   if (response instanceof Response) {
     if (response.headers.has("X-Revalidate"))
-      keys = response.headers.get("X-Revalidate")!.split(",");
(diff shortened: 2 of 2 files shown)


LexSwed:
It's actually my [usage of actions with all the body being on the server](https://github.com/LexSwed/nouvet-solid/blob/7ba371feaf57ed810f350e01419ec6869d6b7643/packages/web/src/server/api/family-invite.ts#L26-L29), makes the response data not `Response`, skipping the check altogether.

https://github.com/solidjs/solid-router/blob/57847624fb673003f6f0d8261ca7352ff00e7480/src/data/action.ts#L150

This makes the situation with server/not server split quite complex to grasp. When running an `action`, `cache` that needs to be revalidated lives on the client. But if all `action` logic is on the server, `handleResponse` can't know which fields need to be invalidated.

LexSwed:
Not fixing the issue I had, it was a different problem.

Issue #422, opened 2024-05-17
A component isActive subpath match incorrect 


madaxen86:
### Describe the bug

"/foo" should not match if current route is "/foobar", but should match for "/foo" or "/foo/...".

### Your Example Website or App

https://stackblitz.com/edit/github-a9ukgq-agqnyl?file=src%2Fapp.tsx

### Steps to Reproduce the Bug or Issue

1. Go to the provided stackblitz
2. click on "A: foo:1"
3. only "A: foo" and "A: foo:1" should be bold
4. The provided Link component has expected behavior. I will provide this solution as a PR. (Note: Difference tu current implementations is that the home route "/" will only be active when the current path is "/".)

### Expected behavior

I expect on route "/foobar" the A with "/foo" not to be active as this is not a subpath of "foo" but just another route the has the same letters at the beginning of the route. 

### Screenshots or Videos

![image](https://github.com/solidjs/solid-router/assets/59310516/823d7599-39ba-476d-9505-0ee86618de06)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

PR #423, opened 2024-05-17
fix: A component isActive subpath matching

madaxen86:
will fix [issue 422](https://github.com/solidjs/solid-router/issues/422)


ryansolid:
Thanks

PR #424, opened 2024-05-20
Mark touchstart event listener passive

Legend-Master:
We don't call `preventDefault` in the handler, so marking it as passive event handler to improve scrolling performance


Legend-Master:
Sorry, touchstart event listeners are passive by default on all modern browsers, please ignore

I probably accidentally clicked Toggle Passive in devtools and it showed me that listener is not passive

PR #427, opened 2024-05-22
Make useCurrentMatches a getter function

### LINKED ISSUE #425: `useCurrentMatches` is not reactive
### Describe the bug

Object returned from `useCurrentMatches` is not reactive at all, wrapping it in createMemo ([as shown](https://github.com/solidjs/solid-router?tab=readme-ov-file#usecurrentmatches) in the router docs) will have no effect. It simply won't get triggered when path/params/info changes. The only way to get this working is to call `useCurrentMatches` again within some effect which is, obviously, not right.

https://stackblitz.com/edit/solid-router-ckudid?file=src/index.tsx

1. Open the repro and your console
2. Start clicking on the links and check out the output
3. `info` will never get updated

### Expected behavior

`useCurrentMatches` should return a proxy that updates whenever a list of matched routes changes.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 125.0.6422.60

### Additional context

_No response_

Brendonovich:
`useCurrentMatches` should return a getter function like `useIsRouting`, `useMatches`, etc.

Closes #425

### DIFF
--- a/README.md
+++ b/README.md
@@ -868,7 +868,7 @@ return <div classList={{ active: Boolean(match()) }} />;
 For example if you stored breadcrumbs on your route definition you could retrieve them like so:
 ```js
 const matches = useCurrentMatches();
-const breadcrumbs = createMemo(() => matches.map(m => m.route.info.breadcrumb))
+const breadcrumbs = createMemo(() => matches().map(m => m.route.info.breadcrumb))
 ```
 
 ### useBeforeLeave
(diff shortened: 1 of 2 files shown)


ryansolid:
Yeah this was probably an oversight.. I was seeing a function calling a function.. yeah ok technically it is breaking. So I guess bump a minor.

ryansolid:
Actually I documented this working the other way so this is a bug..

PR #432, opened 2024-05-29
Export all types used in public API

jpdutoit:
## The problem

Some types referenced in the public API arenot exported. 

This makes it difficult to write code that uses those types, as they cannot be referenced by name.

Specifically: MatchFilter, MatchFilters, PathMatch, Route and RouteMatch

## The fix

Export those types


ryansolid:
Thank you

PR #435, opened 2024-05-30
Fixes #434 - Export type Route as RouteInfo

### LINKED ISSUE #434: In the new version, Route is imported as a type rather than a method.
### Describe the bug

I've set up a new Solid project and then installed the latest version of the router(0.13.4). Although the router is functioning properly, within vscode, Route is being imported as a type instead of a method.
Then, I downgraded the router's version to 0.13.3, restarted vscode, and everything worked as expected. 

sry, i don't have

1. pnpm create solid
2. pnpm add @solidjs/router
3. pnpm install
4. use Router and Route

### Expected behavior

no error in vscode

### Screenshots or Videos

error1
error2

### Platform

- OS: Windows
- Browser: Edge

### Additional context

_No response_

jpdutoit:
Fixes #434 

Unfortunately in #432, the exported type and value `Route` did not merge correctly, breaking the library. It is possible to re-export them under the same name, but since the `Route` type was not part of the public API before renaming it seems cleaner.

I chose `RouteInfo` as the new exported name for the type.

If you prefer, I could rename type `Route` globally in the project instead of just the export.

### DIFF
--- a/.changeset/many-shirts-tie.md
+++ b/.changeset/many-shirts-tie.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": minor
+---
+
+Fix #434 - export type Route as RouteInfo
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -29,7 +29,7 @@ export type {
   RouteLoadFunc,
   RouteLoadFuncArgs,
   RouteDefinition,
-  Route,
+  Route as RouteInfo,  
   RouteMatch,
   RouterIntegration,
   RouterUtils,


ryansolid:
I see it isn't the Route Component but the internal Route representation. It's unfortunate that info already is a field, but I'm having a hard time coming up for a name for this one. We already have `RouteDefinition` for the input, `Route` for the component. Maybe RouteDescription. I see the problem I will try to work with this. Thanks.

PR #437, opened 2024-06-05
Update types.ts

atk:
The useCurrentMatches method is currently mistyped, since router.matches is an accessor. This PR fixes the issue.


Brendonovich:
`useCurrentMatches` is no longer mistyped and [`e.router.matches`](https://github.com/solidjs/solid-router/blob/main/src/routers/components.tsx#L109) has always been an array not an accessor, i think this can be closed

PR #439, opened 2024-06-07
Forward absolute redirects inside `cache` from server to client

### LINKED ISSUE #438: Absolute redirects inside `cache` are ignored if not called in `load`
### Describe the bug

From https://github.com/solidjs/solid-start/issues/1512

```tsx
import { type RouteDefinition } from '@solidjs/router'
import { cache, createAsync, redirect } from '@solidjs/router'
import { getRequestEvent } from 'solid-js/web'

const exampleAction = cache(async () => {
  'use server'
  const location = new URL(getRequestEvent()!.request.url)
  return redirect(`${location.origin}/target`)
}, 'exampleAction')

export const route = {
  load() {
    // redirect will start working if uncommented:
    // void exampleAction()
  },
} satisfies RouteDefinition

export default function Component() {
  const data = createAsync(() => exampleAction())
  return <h1>Result: {data()}</h1>
}
```

> With the route above, when you navigate to /redirect, it does not redirect you to /target.

> Unless you trigger an action with redirect inside route.load or use a relative redirect which starts with / instead of a full URL with the origin, i.e., redirect('/target') works, but redirect('http://localhost:3000/target') does not.

N/A

Use the above code in an SSR route

### Expected behavior

Redirect happens without a load function.

### Screenshots or Videos

_No response_

### Platform

N/A

### Additional context

_No response_

Brendonovich:
Currently, absolute redirects done inside `cache` on the server are (funnily enough) only performed when the cached function is also called inside `load`. I'm not exactly sure how `load` is involved to make that the case, but this PR seems to fix it.

Instead of server-side absolute redirects just being ignored in `handleResponse` by an early return, which results in the redirect Response not being serialized and processed on the client inside `cache`, the early return will only be done if the response has actually been handled. This change results in `load` not having any effect on the redirect behaviour, and absolute redirects will always be passed through from the server to be handled on the client.

Forcing absolute redirects onto the client works fine and avoids any CORS cross-origin redirect problems, but should there be an option for server-side absolute redirects to be properly supported?

Closes #438

### DIFF
--- a/.changeset/breezy-trainers-marry.md
+++ b/.changeset/breezy-trainers-marry.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Forward absolute redirects inside `cache` from server to client
--- a/src/data/cache.ts
+++ b/src/data/cache.ts
@@ -166,21 +166,24 @@ export function cache<T extends (...args: any) => any>(fn: T, name: string): Cac
     function handleResponse(error: boolean) {
       return async (v: any | Response) => {
         if (v instanceof Response) {
-          if (v.headers.has("Location")) {
-            if (navigate) {
+          const url = v.headers.get(LocationHeader);
+
+          if (url !== null) {
+            let returnEarly = true;
+
+            // client + server relative redirect
+            if (navigate && url.startsWith("/"))
               startTransition(() => {
-                let url = (v as Response).headers.get(LocationHeader);
-                if (url && url.startsWith("/")) {
-                  navigate!(url, {
-                    replace: true
-                  });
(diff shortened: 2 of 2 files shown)


ryansolid:
It's weird but I will take it.

Brendonovich:
Yeah copying how `navigateFromRoute` does server redirects seems to work. This does introduce the possibility for some server redirects to get blocked by CORS bc cross-origin, but that's probably something the developer should deal with.

```ts
else if (isServer) {
  const e = getRequestEvent()!;
  e.response = { status: 302, headers: new Headers({ Location: url }) };
  return;
}
```

PR #441, opened 2024-06-08
Ignore non-native LocationChanges in native navigation effect

### LINKED ISSUE #433: Actions called in onMount (or first run effect) are destroyed by start process, and do not receive values nor errors
### Describe the bug

Calling an action (useAction) in onMount is destroyed by a post-hydration process and break their ability to receive values or trap exceptions.

https://discord.com/channels/722131463138705510/910635844119982080/1245271486881857638

and the reproducible case linked:
https://stackblitz.com/edit/github-hzq3du-fznnvu?file=src%2Fcomponents%2FHomeRoute.tsx

(note that it can work the first run or after HMR, so use refresh to see the fail)

This case fails (see console logs).  Work arounds:  
1. a slight delay in timer before executing action.  
2. lazy load route

both might be fragile as not sure timing is deterministic. 

Specific cause mentioned in discord:
https://discord.com/channels/722131463138705510/910635844119982080/1245278570104754228

https://stackblitz.com/edit/github-hzq3du-fznnvu?file=src%2Fcomponents%2FHomeRoute.tsx

1. create action that returns value or throws exception
2. wrap with `useAction` and `useSubmission`
3. call in onMount
4. log in effect the state, value, and error of the action
5. show in JSX results and error state of action

note the logs and display are incorrect, actions goes to undefined.

### Expected behavior

The action works as normal. Values and errors come back to the tracked submission.

### Screenshots or Videos

_No response_

### Platform

All day long, everywhere

### Additional context

_No response_

Brendonovich:
The `createRenderEffect` in charge of resetting submissions when a native navigation occurs currently fires whenever _any_ navigation occurs, including when the router first mounts.
Because of this, when a route first loads or is navigated to that invokes an action in `onMount`, a race condition occurs between the action's submission and the effect's submission reset:
1) The render effect runs (render phase), queueing `submissions[1]([])` onto the next microtask by wrapping it in `start` for a transition.
2) `onMount` runs (post-render phase), invoking the action and creating an entry in `submissions`.
3) The next microtask comes and `submissions[1]([])` runs, wiping out the previously created entry in `submissions`.

I've fixed this by adding an internal `intent` property to `LocationChange`, which if present instructs the render effect to actually run, so it will not run on page load or from `navigate()`. The property has `@internal` so it doesn't show up in the generated `.d.ts`.

Fixes #433

I also ran the `pretty` script bc my local formatter messed some formatting up. If needed I can take just the important changes and get rid of all the formatting changes.


ryansolid:
I have to admit on its own the formatting would have been fine.. but I'm trying to compare this against #442 so not quite clear the differences are. @Brendonovich do you have any thoughts on #442?

Brendonovich:
#422 looks good, it addresses the same things as this PR and some more, I'll close this and suggest some improvements on the other.

PR #442, opened 2024-06-08
Make isRouting more reliable + other fixes

jpdutoit:
I set out to fix #311, but in doing so discovered a few other issues

The isRouting value was not very reliable, and could change many times between true and false

This PR changes `isRouting` to update once to true when routing starts, and then back to false when routing ends - but only *after* the source has been updated. The ordering will allow one to use `isRouting` turning false as a trigger for when it is safe to update your document title. I added tests for these.

While doing this I discovered and fixed the following issues:
- Every navigate causes a "native" transition (As in #441)
- The value of intent is non-deterministic under some conditions

Question:
I noticed that a transition was started on creation. Is that initial transition important? I currently have it commented out.


src/routing.ts
```diff
@@ -301,13 +301,38 @@ export function createRouterContext(
...
+  // Keep track of last intent and target, so that last call to transition wins
+  let lastTransitionIntent: Intent | undefined;
+  let lastTransitionTarget: LocationChange | undefined;
+
+  // Transition the location to a new value
+  const transition = (newIntent: Intent, newTarget: LocationChange) => {
+    intent = lastTransitionIntent = newIntent;
+    lastTransitionTarget = newTarget;
+
+    if (++activeTransitions === 1) setIsRouting(true);
```
Brendonovich, line 315: Can `activeTransition` be replaced by comparing `newTarget` and `lastTransitionTarget`? [In my own testing](https://github.com/jpdutoit/solid-router/compare/is-routing-and-intent...Brendonovich:solid-router:442-improvements) it seems to work

src/routing.ts
```diff
@@ -341,22 +366,17 @@ export function createRouterContext(
...
-      });
-    });
+    const newSource = source();
+    if (newSource.value !== untrack(reference) || newSource.state !== untrack(state)) {
+      transition("native", newSource);
+    }
   });
 
+  // Is this needed?
+  //createRenderEffect(() => transition("initial", untrack(source)));
```
Brendonovich, line 378: nah this isn't needed, `initial` is only useful for loaders and is handled elsewhere

src/routing.ts
```diff
@@ -301,13 +301,38 @@ export function createRouterContext(
...
+
+    if (++activeTransitions === 1) setIsRouting(true);
+
+    startTransition(() =>
+      untrack(() => {
+        setReference(lastTransitionTarget!.value);
+        setState(lastTransitionTarget!.state);
+        resetErrorBoundaries();
+        submissions[1]([]);
+      })
```
Brendonovich, line 323: `untrack` can be removed

src/routing.ts
```diff
@@ -341,22 +366,17 @@ export function createRouterContext(
...
-        resetErrorBoundaries();
-        submissions[1]([]);
-      }).then(() => {
-        intent = undefined;
-      });
-    });
+    const newSource = source();
+    if (newSource.value !== untrack(reference) || newSource.state !== untrack(state)) {
+      transition("native", newSource);
+    }
```
Brendonovich, line 374: using `on(source, ...)` here should remove the need for `untrack`

Brendonovich (approved):
Tested this inside Mattrax and it seems to be working

src/routing.ts
jpdutoit: Yeah, this is cleaner! I pulled in your commits

jpdutoit:
Pulled in your improvements, and added a few minor ones on top

src/data/cache.ts
```diff
@@ -187,7 +187,7 @@ export function cache<T extends (...args: any) => any>(fn: T, name: string): Cac
         return v;
       };
     }
-  }) as CachedFunction<T>;
+  }) as unknown as CachedFunction<T>;
```
jpdutoit, line 190: Not related, but needed for newer typescript versions

ryansolid:
Thanks you both.

ryansolid:
Thanks you both.

PR #446, opened 2024-06-11
Handle absolute redirects within `cache` on server

Brendonovich:
Follows on from #439, handling absolute redirects properly on the server instead of letting them be handled on the client.
This just takes the server redirect implementation in `navigateFromRoute` and puts it in `handleResponse`.


ryansolid:
Thanks

PR #453, opened 2024-06-19
add usePreloadRoute export

kiuKisas:
usePreloadRoute export was missing ,and so it was not possible to import


ryansolid:
Yeah.. sorry.. missed that re-export in my haste.

PR #463, opened 2024-07-23
Support temporary `latest` field for `createAsync` and `createAsyncStore`

elite174:
This PR adds support for `latest` field for `createAsync` and `createAsyncStore` for compatibility with `createResource`.

**MOTIVATION**

At the moment `createAsync` and `createAsyncStore` are just wrappers for `createResource`. `createResource` provides the mechanism to avoid triggering `Suspense`, which is quite useful in some cases.

In order to improve adoption of `createAsync` and `createAsyncStore` it would be useful to provide this mechanism also for these primitives.

In the future major release of solid support of the `latest` field (and `createResource`) will be dropped in favor of `createAsync` and `createAsyncStore`. `.latest` could be replaced with `latest(...)` helper (TBD) or could be dropped at all. In both cases the migration will be simple:
- Replaced with `latest(...)` helper: `data.latest` => `latest(data)`
- Dropped support of `.latest` field: `data.latest` => `data`

At the moment lack of `.latest` support for `createAsync` and `createAsyncStore` prevents people from wide adoption of these primitives. Moreover, people start invent some hacks like [this](https://discord.com/channels/722131463138705510/722131463889223772/1264787694614085683) which consume much more memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,11 +495,19 @@ This is light wrapper over `createResource` that aims to serve as stand-in for a
 const user = createAsync((currentValue) => getUser(params.id))
 ```
 
+It also preserves `latest` field from `createResource`. Note that it will be removed in the future.
+
+```jsx
+const user = createAsync((currentValue) => getUser(params.id))
+return <h1>{user.latest.name}</h1>;
+```
+
 Using `cache` in `createResource` directly won't work properly as the fetcher is not reactive and it won't invalidate properly.
 
 ### `createAsyncStore`
 
 Similar to `createAsync` except it uses a deeply reactive store. Perfect for applying fine-grained changes to large model data that updates.
+It also supports `latest` field which will be removed in the future.
 
 ```jsx
 const todos = createAsyncStore(() => getTodos());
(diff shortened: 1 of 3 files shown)


ryansolid:
Yeah.. I was hoping to avoid this until we had a better idea of what the future API would look like. But it seems we're here. Having something like `.latest` is relatively low risk as you said as it isn't transitive and it doesn't cause poor patterns like `.state` or `.error` or `.loading` can unintentionally. So my gut is we can move with this and the migration cost will be relatively minor. It will still be something, but it is localized mostly.

elite174:
Could someone merge this thing then?

elite174:
@ryansolid Could someone merge this PR please, if you agree?

ryansolid:
I'm not ignoring this. I am just being a little slow. We need to be careful and I'm actively looking at stuff right now that I figured a few more days wouldn't make a huge difference. If I don't get to where I want by the end of the week we can go ahead.

ryansolid:
I write about my thinking here: https://dev.to/this-is-learning/async-derivations-in-reactivity-ec5

PR #465, opened 2024-08-07
feat: Wire useSearchParams to url rewriting

jchatard:
This is a follow up on [feat: add URL transformation hook](https://github.com/solidjs/solid-router/pull/410)

My implementation of URL rewriting was working great until I tried using `useSearchParams` setter which would update the browser's URL to the route:

From: `/foo.html?bar=baz` updating `bar` to `biz`
Result: `/product/123?bar=biz`
Expected: `/foo.html?bar=biz`

So I needed  to wire it, but the logic is reversed, compared to the previous PR.

- `transformUrl` transforms incoming requests pathname to router's route.
- for `useSearchParams` setter we the need the opposite, from a route displaying the rewritten url in the browser.

So to make this reversed behavior clearer, I renamed `transformUrl` (with a deprecation notice, even though I'm pretty sure I'm the only one using rewrites :-) ) to `rewritePathToRoute` and introduced `rewriteRouteToPath`.

I plan to write some doc at some point, but I might be still missing some rewrite wiring here and there.

And you implement it like this:

```tsx
const pathnameToRoute = (url: string) => {
    const u = new URL(url, "http://localhost/");
    return rewriteMap[u.pathname] ? rewriteMap[u.pathname] + u.search : url;
  };

  const routeToPathname = (url: string) => {
    const u = new URL(url, "http://localhost/");
    if (Object.values(rewriteMap).includes(url)) {
      const swapped = swapKeyValue(rewriteMap);
      return swapped[u.pathname] + u.search;
    }
    return url;
  };

  return (
    <Router
      explicitLinks={true}
      preload={false}
      rewritePathToName={pathnameToRoute}
      rewriteRouteToPath={routeToPathname}
      root={(props) => (
        <Suspense>
          <RootContexts>{props.children}</RootContexts>
        </Suspense>
      )}
      rootPreload={() => {
        getMenuMain();
        getMenuFooter();
        getSettings();
      }}
    >
      <FileRoutes />
    </Router>
  );
```

Let me know what you think!

Thank you!
Jérémy

### DIFF
--- a/.changeset/tricky-rats-raise.md
+++ b/.changeset/tricky-rats-raise.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+feat: Wire useSearchParams to url rewriting
--- a/src/data/events.ts
+++ b/src/data/events.ts
@@ -8,7 +8,7 @@ export function setupNativeEvents(
   preload = true,
   explicitLinks = false,
   actionBase = "/_server",
-  transformUrl?: (url: string) => string
+  rewritePathToRoute?: (url: string) => string
 ) {
   return (router: RouterContext) => {
     const basePath = router.base.path();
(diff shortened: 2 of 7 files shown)


ryansolid:
Hmm.. I wonder if this is necessary. I was hoping we'd only need the transform in one direction. I see it is because we call navigate and we pass in the path there but if there was a way to communicate the params change without including the path I think it would be unnecessary. Of course if there were the whole setter side of `useSearchParams` might be a questionable API. I wasn't the originator of this API and in my previous routers you'd just pass an object to navigate if you wanted to update the search params.

I feel like this is an area that could be streamlined. Given the get is just a projection of location and the set could be handled by navigate. There might be reasoning I'm missing but my gut is we should look at alternatives.

jchatard:
ok, I agree with you.

I just had a look, I'll create a new PR during the day with way more simple.

PR #170, opened 2022-09-20
Setting search params does not change route

jeski-bright:
Use the current location as a base when calling `setSearchParams`.

This doesn't fix the `from` route param though, however it might be desirable to know where the change was triggered.

Fixes #157

PR #466, opened 2024-08-08
feat: Make useSearchParams work with URL rewriting on the client

jchatard:
This is a much simpler implementation of [feat: Wire useSearchParams to url rewriting](https://github.com/solidjs/solid-router/pull/465).

- It doesn't need a second hook to reverse the mapping.
- It doesn't touch the ```navigate``` logic, which, to be honest, I don't fully grasp.

It might be a bit less agnostic, but I can't find a scenario where this could break, given the purpose of the ```useSearchParams``` setter.

I guess a cleaner solution might be to move this to the integration level...

In React Router, they seem to allow ```navigate()``` to take a string like this: ```?foo=bar``` and then append it to the current location pathname. But I'm not sure, as I have never used it.

### DIFF
--- a/.changeset/late-buses-melt.md
+++ b/.changeset/late-buses-melt.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Make useSearchParams work with URL rewriting on the client.
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -104,7 +104,7 @@ export const useSearchParams = <T extends Params>(): [
   const navigate = useNavigate();
   const setSearchParams = (params: SetParams, options?: Partial<NavigateOptions>) => {
     const searchString = untrack(
-      () => location.pathname + mergeSearchString(location.search, params) + location.hash
+      () => isServer ? location.pathname : window.location.pathname + mergeSearchString(location.search, params) + location.hash
     );
     navigate(searchString, {
       scroll: false,


ryansolid:
Sorry thought I replied to this one already.. I must have lost it. In any case I wonder if the API for us is all wrong. If useNavigate could handle query param navigation naturally by not trying to also change the path could doing it both directions be avoided. Like if we could tell it to only apply the params then it wouldn't try to apply the path. Do you think that would work? Am I missing another case where we need to go the other direction?

I had this sort of API in my routers in the past before this one, and wasn't too heavily involved early days with this addition. It seems that query params have been a source of confusion and maybe the problems there rather than in your rewriting.

jchatard:
Yeah that would perfectly work for my use case. In fact, when I first used `useSearchParams`'s setter, I have surprised to see the pathname change as well. So yeah!

ryansolid:
Ok.. that will take a minor bump if we are going to change this I need to review a few things because I'm pretty sure I asked the reasoning when we first added the feature and want to make sure everything is good. https://github.com/solidjs/solid-router/pull/52
I'm pretty sure this conversation happened in discord or over a call because no details here. But my gut is given that our location hook has query there could be a path that gets us away from having this hook at all.

jchatard:
@ryansolid might be https://github.com/solidjs/solid-router/pull/170/commits/23177da55be6ecb2e81bd5cf3dd7b24e2b345241

ryansolid:
> @ryansolid might be [23177da](https://github.com/solidjs/solid-router/commit/23177da55be6ecb2e81bd5cf3dd7b24e2b345241)

I think you are right.. I am going to look at doing a different fix for this issue and we might be all good. I don't think we should hardcode the base path here.. it should be not necessary to do so.

ryansolid:
Oh sorry now I get my confusion. This is a different PR. I did respond to the other one. Upon looking closer the problem is that we use the resolved string to do the comparison. Resolving based on nesting doesn't make sense. It needs to be the global location. Of course with rewriting this is difficult because our representation of location is different than the output one.

Technically I think we could have the React Router behavior but the fix internally is going to be like this. Let me play a bit with it and see if I find something I like.

ryansolid:
Yeah I think https://github.com/solidjs/solid-router/commit/86c3b1f7015c1b7a7dbe95bde37aa4e961bb06ce Get us where we want. There is a bit of an open question about how to handle `useResolvedRoute` because there might be inconsistency there but internally like in the <A> we strip this stuff anyway.

jchatard:
Awesome, I’m going to try this hopefully today and get back to you!

Thanks.

jchatard:
@ryansolid you might want to revert your commit, it breaks useSearchParams I think. Here's a video attached, after I applied your patch to my project.

https://github.com/user-attachments/assets/96d515f7-5fc0-4c07-9da7-0b58db133ea7

jchatard:
Hey @ryansolid I just setup a reproduction based on Solid router's `main` branch with a Cypress E2E to help you out.

Check the README for install: https://github.com/Breek/solid-router-tests

ryansolid:
Thanks you are right.. I thought the problem was somewhere it wasn't. The value is already transformed so it isn't an issue. Except I did notice a discrepancy.. we transform back on browser operations but not on internal nav. This causes a lot of weirdness. I'm not sure which side it is supposed to be but right now (pre any fix here) the main solution for transforms is inconsistent. 

Basically on initial load it sets value to `/product/:id` etc.. and then on navigation it sets it to the transformed value. So I removed rawPath had a fix that worked except on first navigation. Either we should be transforming back earlier always or I think we should remove the transform from the integration section.

jchatard:
@ryansolid it's too late here for my brain to answer :-). I'll get back to you tomorrow!

Thank you!

jchatard:
Thanks @ryansolid  I tested, and it works great.

Regarding `useLocation()` it seems to me (but this is very personal) that it makes sense that it matches what the browser's URL looks like. So I'd say we're good.

We probaly can close this now?

ryansolid:
Yeah we are good then I will make the release. 

My only thinking of doing it the other way is that you could basically write your code without knowing about the mapping and have it all work. But I suspect people use stuff like the location path to represent the browser like you said.

PR #477, opened 2024-09-14
feat: Add customSetup for routers to override "NativeEvents"

chiefcll:
Hey @ryansolid - For devices that aren't web based and using universal renderer - I'd like to have a customSetup function to override the default "nativeSetup" (as we don't use a mouse pointer). Just want to get your initial feedback before I do it to the other routers as well.

### DIFF
--- a/src/routers/HashRouter.ts
+++ b/src/routers/HashRouter.ts
@@ -40,7 +40,7 @@ export function HashRouter(props: HashRouterProps): JSX.Element {
           delta => !beforeLeave.confirm(delta && delta < 0 ? delta : getSource())
         )
       ),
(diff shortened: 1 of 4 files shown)


ryansolid:
Hmm.. I was of the mind that you'd just make a different router using `createRouter` rather than add more options to the in built one. Basically a step higher in the process. Like the `create` hook already exists. These routers assume a DOM environment another might not.

It is interesting if this is the only change but you can see I was thinking there could be other differences.

chiefcll:
Gotcha... I have a Browser environment, just not a HTML DOM environment. So I'll need everything else in the HashRouter...

chiefcll:
I'll go that route and just use createRouter in my app. thanks!

Issue #328, opened 2023-05-28
`useSearchParams` suggestion


rattleSSnake:
It would be very needed that `setSearchParams` from *solid-start* works like setting state, meaning it holds previous value and overwrites existing state when omitting prev. That way, it both works when you want to overwrite existing or append.
```js
setSearchParams({ hello: "world" }) //overwrites existing
setSearchParams(prev => ({...prev, hello: "world" })) //appends
```

chiefcll:
Just to add in to this - I'm looking to support Browsers before Proxy support is added. I'd prefer for it also to be a plain Object. If there is any way to do this without proxy (maybe defining which search or query params one was looking for). I'll be diving into Router code this week probably forking it to remove that for now as I don't need reactive params.

ryansolid:
> Just to add in to this - I'm looking to support Browsers before Proxy support is added. I'd prefer for it also to be a plain Object. If there is any way to do this without proxy (maybe defining which search or query params one was looking for). I'll be diving into Router code this week probably forking it to remove that for now as I don't need reactive params.

It's tricky from a general sense because params can change shape, ie be added removed. My gut is I can't just get rid of the proxy here.

chiefcll:
I'm looking at this now. I think for params you could process all the routes and look for all the defined params and create an object with get / set for each of the possible params from the routes. 

For Query params you'd need to be told all the possible query params. I think it's doable to avoid the proxy.

PR #478, opened 2024-09-15
fix: Use non proxy params when Proxy isn't available (#328)

chiefcll:
For older devices which don't support Proxy this offers a solution by having params calculated at runtime so a proxy isn't needed. Something similar could be done with query params if user provides the param keys they will use.

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -37,6 +37,7 @@ import type {
 import {
   mockBase,
   createMemoObject,
+  createMemoWithoutProxy,
   extractSearchParams,
   invariant,
   resolvePath,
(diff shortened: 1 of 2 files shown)


ryansolid:
Hmm.. this change I need to think more about. It does add more than a trivial amount of code to support the non-proxy case. I want to consider what we are doing here in general and what can be effectively handled/pulled out.

chiefcll:
Thanks! Yea you could have a simpler / cleaner solution.

ryansolid:
Honestly your solution is quite nice for this. I actually didn't think it was possible to do this without proxies but using the route definition is smart. I just don't want people to pay the price for this generally speaking so thinking of how to make this treeshakable.

Generally it can't just be a toggle or check. The code needs to be passed into the router. Like a custom params handler thing. Then even your custom router could implement it if wanted to.

something like `paramsWrapper`

And then the logic would be:
```js
const params = options.paramsWrapper ?
    options.paramsWrapper(paramsHandler, branches) :
    createMemoObject(paramsHandler);
```

Then you could pass in a paramsWrapper that had the custom code you created. Do you think something like that would work?

Alternatively we could ship with the proxy-less wrapper as a convenience but people would have to manually pass it in.

chiefcll:
Just to play devils advocate. Why not remove Proxy? If query params required defined keys to look for you could do the same thing, and it too wouldn't require a Proxy at that point... Users just pass in queryKeys = [id, type, search] etc.

I'm ok with the paramsWrapper. Can we do similar thing with Query?

ryansolid:
> I'm ok with the paramsWrapper. Can we do similar thing with Query?

Yeah and you'd need that I think because we don't define the query params as part of the route definitions so there would be no way to avoid a proxy in that case unless you defined them somehow which giving you the wrapper gives you the freedom to.

ryansolid:
Yeah I can add both through the Router utils object. This way they aren't necessary to set through the runtime for the end user but when you create your custom router it can have this behavior built in. I think that is probably the right API call. I don't see consumers messing with this as much as custom Router authors.

ryansolid:
Ok released 0.14.8.. Let me know if you need anything else.

PR #481, opened 2024-09-24
Fix scrollToHash to handle hashes starting with a number

CirnoV:
When hash starts with a number, `document.querySelector` throws an error, causing `scrollToHash` to fail.

### DIFF
--- a/.changeset/hungry-trees-explode.md
+++ b/.changeset/hungry-trees-explode.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Fix scrollToHash to handle hashes starting with a number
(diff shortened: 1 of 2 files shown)


ryansolid:
Do you see any reason why we wouldn't always use get document by id? I admit I wasn't responsible for the hash router implementation so the fact that there is all this query selector stuff just for this case sort of surprises me. I will do some forensics here but the native DOM behavior seems to be restricted to ids so I don't see benefit in applying a general selector and trying to restrict it to fit.

CirnoV:
I prefer to use `document.getElementById`, but since there might be cases in the existing project where a selector is used (even though it's not documented), I was a bit hesitant to change the code.

PR #484, opened 2024-10-01
fix active detection for url's containing international characters

rvlzzr:
The active/inactive classes don't apply correctly for URL's containing international characters, as the location is URL-encoded while the path passed to the component is not. This is one way to make it work, the alternatives being to encode the path coming in, or to decode the location everywhere (which I would personally like to see but would be a larger change). Not sure which is preferred.


ryansolid:
I'm not sure which is prefered either. I feel like we tried some encode/decode stuff previously and it got us in trouble. But I am definitely open to finding the right solution. This is easy enough for now but I do wonder if we will find more issues unless we solve this systematically.

PR #487, opened 2024-10-06
fix: add missing exports to recreate Routers

chiefcll:
When trying to tweak HashRouter there were some missing exports. Adding those in.

### DIFF
--- a/src/routers/index.ts
+++ b/src/routers/index.ts
@@ -1,6 +1,6 @@
 export { Route } from "./components.jsx";
 export type { BaseRouterProps, RouteProps } from "./components.jsx";
-export { createRouter } from "./createRouter.js";
+export { createRouter, bindEvent, scrollToHash } from "./createRouter.js";
 export { Router } from "./Router.js";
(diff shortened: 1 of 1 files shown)


ryansolid:
Hmm.. let me see.. these are utility functions that basically have very minimal functionality and I don't know if I want to just expose as public API. I'm going to merge this because I'm committed to making this work easily but I might move some of these internals around a bit before publishing.

chiefcll:
I'm fine with you closing this out too - its not a blocker for me. I only used the bindEvent and it is a small util function, which I just copied from the source. Will let you ponder 👍

ryansolid:
> I'm fine with you closing this out too - its not a blocker for me. I only used the bindEvent and it is a small util function, which I just copied from the source. Will let you ponder 👍

Oh hmm. Well bindEvent was specifically the one I was thinking of that was awkward because it doesn't really do anything outside of this. It was a convenience to share between the routers but was probably unnecessary. `scrollToHash` at least has functionality to it that could be useful somewhere. My gut is to just close it out in this case.

PR #489, opened 2024-10-12
feat(search-params): Enable arrays in useSearchParams

markojerkic:
# The problem

I want to be able to add an array of values as a search parameter. It is supported using native [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) with the method `append`. It is also supported in [React Router](https://reactrouter.com/en/main/hooks/use-search-params) (you can see that in the type definition, and I've used it multiple times.).

## Desired behaviour

```ts
setSearchParams({
    foo: ["bar", "bazz"]
})
```
This would result in `?foo=bar&foo=bazz`

# Proposed solution

In serialization check if input is array, then use the `append` method, else use the `set` method.
https://github.com/solidjs/solid-router/blob/b4f75446f802f6e11d5e07b8beb1e6bb9e096582/src/utils.ts#L172-L178
In deserialization, check if the key is already present. If yes, then transform the value into an array and append to it.
https://github.com/solidjs/solid-router/blob/b4f75446f802f6e11d5e07b8beb1e6bb9e096582/src/utils.ts#L50-L56

I also wrote a couple of tests to verify the behaviour.

### DIFF
--- a/src/routing.ts
+++ b/src/routing.ts
@@ -32,7 +32,9 @@ import type {
   RouterContext,
   RouterIntegration,
   SetParams,
-  Submission
+  Submission,
+  SearchParams,
+  SetSearchParams
 } from "./types.js";
 import {
   mockBase,
@@ -96,13 +98,13 @@ export const useCurrentMatches = () => useRouter().matches;
 
 export const useParams = <T extends Params>() => useRouter().params as T;
 
-export const useSearchParams = <T extends Params>(): [
+export const useSearchParams = <T extends SearchParams>(): [
   Partial<T>,
-  (params: SetParams, options?: Partial<NavigateOptions>) => void
+  (params: SetSearchParams, options?: Partial<NavigateOptions>) => void
 ] => {
   const location = useLocation();
   const navigate = useNavigate();
-  const setSearchParams = (params: SetParams, options?: Partial<NavigateOptions>) => {
+  const setSearchParams = (params: SetSearchParams, options?: Partial<NavigateOptions>) => {
     const searchString = untrack(() => mergeSearchString(location.search, params) + location.hash);
     navigate(searchString, {
       scroll: false,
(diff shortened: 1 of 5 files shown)


ryansolid:
I want to think about the serialization format. Having the same named key is one way. Another would be to JSON serialization + uriEncoding. I suppose you could do that today by setting JSON serialized value and parsing it. I guess the format presented here could only be done internally.

Hmm.. I wonder if there is other value to handle search params internally. I think that might inform our solution here. I admit I am not up to date with the state of the art here and probably need to take some time to look at Tanstack Router.

markojerkic:
Tanstack Router keeps the `URLSearchParams` encoding for the first level of keys. The second level of depth gets JSON serialized + URI encoded. Arrays too are JSON serialized. 

I guess, since the current router supports only `Record<string, string>` as search parameters, neither solution would really be a breaking change.

Although the JSON approach means an expanded API, it does potentially create some headache. For example, the merge state algorithm. With the flat single-level design, it is very easy to merge state with an existing state - `{...prevState, ...newState}`. But if we were to add nesting to the equation, it might be a bit more difficult. 

The `createStore` API might be useful to reuse here. But that would probably mean a breaking change, since with current API `setSearchParams({foo: "bar"})` would add the `foo=barr` to the existing parameters creating for example `?foo=bar&bazz=xyz`, and the store approach would override the current state (unless you would use the function setter with the previous value).

ryansolid:
Hmm.. yeah I was trying to avoid Stores as a necessary dep here. That being said because browser native navigation we have a situation of needing to serialize/deserialize back and forth which would require diffing. My gut here is we should treat each param as atomic with no potential for going deeper from a change detection perspective. 

Even if we were to deeper serialize data there would be no expectation for anything but the key to trigger. Ie.. merging would always be shallow. I was thinking that TypeScript might be a reason to be more declarative with props but without back linking we probably would get no benefit here.

I suppose the encoding format doesn't matter all that much. Just the behavior. As long as you can set and retrieve array or object values and it works then you are fine. The fact that the native `URLSearchParams` encodes it the way you are proposing I'm leaning more towards it.

ryansolid:
Thank you

PR #490, opened 2024-10-15
rename `cache` to `query`, action `onComplete`

ryansolid:
Proposed updates to Router APIs...

Mainly rename `cache` to `query`

Also add the `onComplete` API to submissions. That I brainstormed here: https://hackmd.io/@0u1u3zEAQAO0iYWVAStEvw/ry9vVctJkg

### DIFF
--- a/src/data/action.ts
+++ b/src/data/action.ts
@@ -3,7 +3,7 @@ import { isServer } from "solid-js/web";
 import { useRouter } from "../routing.js";
 import type { RouterContext, Submission, SubmissionStub, Navigator, NarrowResponse } from "../types.js";
 import { mockBase } from "../utils.js";
-import { cacheKeyOp, hashKey, revalidate, cache } from "./cache.js";
+import { cacheKeyOp, hashKey, revalidate, query } from "./query.js";
 
 export type Action<T extends Array<any>, U, V = T> = (T extends [FormData] | []
   ? JSX.SerializableAttributeValue
@@ -62,7 +62,15 @@ export function useAction<T extends Array<any>, U, V>(action: Action<T, U, V>) {
 export function action<T extends Array<any>, U = void>(
   fn: (...args: T) => Promise<U>,
   name?: string
-): Action<T, U, T> {
+): Action<T, U>
+export function action<T extends Array<any>, U = void>(
+  fn: (...args: T) => Promise<U>,
+  options?: { name?: string; onComplete?: (s: Submission<T, U>) => void }
+): Action<T, U>
+export function action<T extends Array<any>, U = void>(
+  fn: (...args: T) => Promise<U>,
+  options: string | { name?: string; onComplete?: (s: Submission<T, U>) => void } = {}
+): Action<T, U> {
   function mutate(this: { r: RouterContext; f?: HTMLFormElement }, ...variables: T) {
     const router = this.r;
     const form = this.f;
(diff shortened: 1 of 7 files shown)


atilafassina:
Instead of completely replacing on this release, how about we add a deprecation warning so the minor doesn't break existing apps?

```ts
/**
 * @deprecated Use `query()` instead..
 * This method will be removed on v0.15.0+
 */
export const cache = query
```

This will allow TS codebases at least to have a clear distinct warning on their code:


Screenshot 2024-10-17 at 09 01 42

davedbase:
I agree with @atilafassina here. Let's keep `cache` available with a deprecation notice and provide `query` as the new option. In a few releases we can officially deprecate it, just give users some warning before we do.

brenelz:
If i'm not mistaken I think he did this already here?
https://github.com/solidjs/solid-router/pull/490/files#diff-ec90fda2df5f2c5fb4fdc9858fdece2ed80173595f5443fd20720df0a6623b12R233

src/data/action.ts
```diff
@@ -76,6 +84,7 @@ export function action<T extends Array<any>, U = void>(
     function handler(error?: boolean) {
       return async (res: any) => {
         const result = await handleResponse(res, error, router.navigatorFactory());
+        o.onComplete && o.onComplete(submission);
```
brenelz, line 87: So I have already been playing with this `onComplete` hook a bit. Bleeding edge I know lol

I think this should actually be moved down a couple lines to be after you've `setResult(result)`. Otherwise the submission.result is always undefined in the `onComplete` handler

ryansolid: Good call.. Hmm.. it's tricky because when we set the result we trigger reactivity. I guess I should make a copy of the submission object that overrides that property value. I think timing wise we want to run `onComplete` before we actually render anything.

devagrawal09:
I really like the new action api to allow for those advanced optimizations for people that are willing to put effort into manual cache management.

My only slight disappointment is that if I'm writing manual cache mutations anyways, I would like it to optimize all the way and prevent any reconciliation by making the cache update fine grained. This is still a good optimization over the default since it saves server resources and latency, but it feels like there is potential to save on data reconciliation as well without adding any more work for the developer.

What if `query` could be configured to use a store under the hood, and `createAsyncStore` can check if the thing I returned was a store and just let it through?

ryansolid:
@devagrawal09 I've thought about this before. As you can imagine I moved stuff all around originally. The Store on the inside is awkward for hydration and serialization. It needs to upgrade to a store without running the store code on the client. There may be solutions there but finding a good solution for this with the constraints of how createResource works is pretty tough.

devagrawal09:
Aww hydration PITA strikes again

So am I hearing that this idea could be revisited in 2.0 once createAsync is the building block?

ryansolid:
> Aww hydration PITA strikes again
> 
> So am I hearing that this idea could be revisited in 2.0 once createAsync is the building block?

Maybe but it also brings with it the question of if createAsync should automatically serialize for hydration. That's more of the challenge here. We could make a change where it doesn't. Which would make these sort of things easier but we'd need to still find ways for people not using this stuff to find success.

src/data/action.ts
ryansolid: Ok I think I solved it with the latest version I just pushed. We make a clone and then intercept the retry operation to ensure that it works as expected.

atilafassina:
Docs are ready:

https://github.com/solidjs/solid-docs/pull/937

PR #495, opened 2024-11-08
Add JSdoc

marvin-j97:
As of now, the IDE experience with Solid Router is subpar because of missing JSdocs. I added JSdocs for most Router functions:

### Before
image

### After
image


ryansolid:
Looks great. Thank you.

PR #498, opened 2024-11-27
Vite 6 compat

birkskyum:
Related to:
- https://github.com/solidjs/vite-plugin-solid/pull/166
- https://github.com/solidjs/templates/pull/153
- https://github.com/vitejs/vite/pull/18792

### This PR

This PR updates all the deps to their latest, including the upgrade to Vite 6.

The only migration change needed is the resolve.conditions: https://vite.dev/guide/migration#default-value-for-resolve-conditions

I noticed there wasn't test CI, which I found unusual, so I added that in another PR:
- #499


ryansolid:
I don't think this is minor because it vite only impacts the building of this repo. But I'm good with all the changes. So I will adjust accordingly.

PR #499, opened 2024-11-27
Add github test workflow

birkskyum:
pnpm install, build, test

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,30 @@
+name: Tests
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


atilafassina:
hey, @birkskyum . Maybe stupid question but: do we really need `pnpm run build` in this CI?
I only see Vitest files in this workflow and I think they pull files from `/src`

birkskyum:
Good catch, it's not needed. Ran i locally just now without the build to make sure, and it's all green.

ryansolid:
Thank you

PR #508, opened 2025-01-15
export SearchParams

AlexErrant:
`SearchParams` is on `Location` which is already exported. This PR makes it so I don't have to do `Location['query']`, e.g.

```ts
import { type Location } from '@solidjs/router'

export const someFn = (search: Location['query']) => { ... }
```

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -19,6 +19,7 @@ export * from "./data/index.js";
 export type {
   Location,
   LocationChange,
+  SearchParams,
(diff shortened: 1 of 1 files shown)


ngdangtu-vn:
Just what I plan to do, thanks mate.

ngdangtu-vn (approved):

birkskyum:
@AlexErrant can you add a changeset? `pnpm changeset`

PR #531, opened 2025-06-10
fix: createAsync - catch errors of prev to avoid bubbling error up

madaxen86:
currently when the inner function of createAsync's inner function throws after an arg has updated, the error will not be caught by the closest ErrorBoundary that wraps the accessor. It will bubble up instead. Which is kind of inconsistent, because if there is an initial error it will be caught by the closest ErrorBoundary

wrapping the the prev in catchError and returning undefined instead resolves this.

```ts
[resource] = createResource(
    () =>
      subFetch(
        fn,
        catchError(
          () => untrack(prev),
          () => undefined
        )
      ),
    v => v,
```
Also added tests for a few different simple scenarios of create async.

[Simple reproduction](https://playground.solidjs.com/anonymous/1744fd48-1bb4-49dd-8220-797a0194c40b)

### DIFF
--- a/src/data/createAsync.ts
+++ b/src/data/createAsync.ts
@@ -1,7 +1,14 @@
 /**
  * This is mock of the eventual Solid 2.0 primitive. It is not fully featured.
  */
-import { type Accessor, createResource, sharedConfig, type Setter, untrack } from "solid-js";
+import {
+  type Accessor,
+  createResource,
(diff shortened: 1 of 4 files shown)


ryansolid:
Thank you. There are definitely some gaps because the way we hacked around createResource.

PR #554, opened 2026-03-03
feat: Add `batchedQuery`

lxsmnsyc:
## Example

```js
const sleep = (value, ms) => new Promise((res) => setTimeout(res, ms, value));

const example = batchedQuery(async (queries) => {
  const map = new Map();

  for (const query of queries) {
    map.set(query, `hello, ${query}`);
  }

  return await sleep(map, 500);
}, (data, query) => data.get(query));

async function test() {
  const results = await Promise.all([
    example('foo'),
    example('bar'),
  ]);
  console.log(results);
}
```

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,0 +1,2 @@
+onlyBuiltDependencies:
+  - esbuild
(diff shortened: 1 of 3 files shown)


lxsmnsyc:
Closing in favor of #607

PR #563, opened 2026-05-25
ci: declare workflow-level `contents: read` on test and dist-typecheck

arpitjain099:
Both workflows run pure tests / typecheck. No GitHub API writes, so workflow-level `contents: read` is the right cap.

Post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.


ryansolid:
Thank you.

Issue #542, opened 2025-09-08
Revalidating breaks form actions


kf5v:
### Describe the bug

When revalidation is triggered, forms that have a defined action lose their connection with the  solid-router.

The form submission bypasses the router and redirects the browser to a URL like:
`http://localhost:3000/_server?id=src_routes_about_tsx--helloWorld_action&name=....`

### Your Example Website or App

https://stackblitz.com/edit/github-p2jyupw6?file=src%2Froutes%2Fabout.tsx

### Steps to Reproduce the Bug or Issue

1. On About page
2. Click any "Hello World!" button to submit the form once
3. The form is submitted and `items` get revalidated
4. Now when you try to submit once gain any of the form - application is redirected to white page

### Expected behavior

I would expect that after revalidation any forms with actions will continue to be interactive with solid-router

### Platform

- OS: Linux
- Browser: Chrome
- Version: 123

### Additional context

If instead of array of objects, `getItems`  returns array of string
so instead of
```
  return [...'abcdefghijklmnopqrstuvwxyz'].map((name) => ({ name }));
```
it is
```
  return [...'abcdefghijklmnopqrstuvwxyz'];
```
and later using the string directly
then the issue does not appear

ryansolid:
Fixed on main in cae1d15. When revalidation re-created a bound action under the same URL, the disposal of the older owner deleted the newer registration from the actions map, so subsequent submits fell through to native form submission at `https://action/...`. Cleanup is now guarded to only remove a registration it still owns. (The equivalent fix for the 2.0/next branch is in #568.) Will be in the next patch release - please reopen with the repro if it still occurs there.

PR #568, opened 2026-07-10
Fix bound action registration cleanup

brenelz:
## Summary

- preserve a newer bound action registration when an older owner with the same action URL is disposed
- add regression coverage for repeated `.with()` bindings that share a URL
- add a patch changeset

This fixes forms falling through to native submission at `https://action/...` after an optimistic update recreates a bound action.

## Testing

- `pnpm test`
- `pnpm vitest run test/data/action.spec.ts test/data/events.spec.ts`
- `pnpm test:types`

### DIFF
--- a/.changeset/quiet-actions-stay.md
+++ b/.changeset/quiet-actions-stay.md
@@ -1,0 +1,5 @@
+---
+"@solidjs/router": patch
+---
+
+Prevent stale bound action cleanup from unregistering a newer action with the same URL.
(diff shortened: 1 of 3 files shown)


ryansolid:
Thanks @brenelz — good catch, and you landed on exactly the right fix. The same guard shipped in 0.16.2 on `main` as part of the recent bug-fix batch (#542) and was merged forward to `next` in 996e8e5: `onCleanup(() => actions.get(url) === fn && actions.delete(url))`, with equivalent regression coverage in `test/data/action.spec.ts` ("disposing an older owner should not unregister a newer action with the same URL"). Closing since it's already fixed on both branches.

PR #573, opened 2026-07-22
Fixes Router.paths infered types.

m-canton:
Fixes error that appears when Router.paths type is infered:

> The inferred type of 'paths' cannot be named without a reference to 'DefaultSearchTypes' from '../node_modules/@solidjs/router/dist/paths'.


ryansolid:
Thanks @m-canton! Verified the fix reproduces and resolves TS2742 in a consumer project emitting declarations for an inferred `Router.paths` — including nested routes, params, and match filters (no other internal types leak). Merged, and it'll go out in the next prerelease.

PR #580, opened 2026-07-27
Calls Action.onSettled on null responses

m-canton:
It calls `Action.onSettled` on any response is needed to clear states changed by `Action.onSubmit` or before. Only redirects are excluded.

For example, it allows to close form dialogs on action completed.

### DIFF
--- a/src/data/action.ts
+++ b/src/data/action.ts
@@ -412,7 +412,7 @@ let flightApplications = 0;
 export function setupFlightDataConsumer(router: RouterContext) {
   return subscribeFlightData<Record<string, any>>((data, { response }) => {
     flightApplications++;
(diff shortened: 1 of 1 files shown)


ryansolid:
Fixed on `next` in e1c1ec4. Every invocation now settles its hooks exactly once — void mutations, metadata-only responses, and redirects included. We went one further than proposed: redirects also fire `onSettled`, so the contract is simply "one `onSettled` per invocation, no exceptions" (skipping it on redirect risks leaving stale `onSubmit` state — like an open dialog — sitting over the new page, since a navigation doesn't necessarily unmount UI rendered above the route tree).

The submissions book-keeping is unchanged and intentional: only outcomes with a result or an error enter the list, so the typical void mutation still leaves nothing behind for you to clear.

PR #584, opened 2026-08-09
perf(A): cut server render cost by ~5x

### LINKED ISSUE #583: <A> costs ~6us of server CPU per instance during SSR (20x a plain <a>), mostly mergeProps/splitProps
## Summary

`<A>` costs roughly **6 µs of server CPU per instance** during SSR, about **20x** a plain `<a>`, and about 40% of that is `mergeProps` + `splitProps` building accessor-backed prop objects that can never be observed on the server, since `renderToString` resolves everything exactly once.

On link-dense server-rendered pages this dominates the response. A 1000-link table renders in 0.31 ms with plain `<a>` and 6.29 ms with `<A>`.

Versions: `@solidjs/router@1.0.0`, `solid-js@1.9.14`. The same hot path is present in `0.15.4`; I diffed `A` across the two and it is unchanged apart from a trailing-slash tweak in `isActive`.

## Reproduction

Self-contained, no framework needed.

`package.json`
```json
{
  "name": "solid-router-a-ssr-repro",
  "private": true,
  "type": "module",
  "scripts": { "bench": "vite build --ssr bench.jsx --outDir dist && node dist/bench.js" },
  "dependencies": { "@solidjs/router": "1.0.0", "solid-js": "1.9.14" },
  "devDependencies": { "vite": "^7.0.0", "vite-plugin-solid": "^2.11.0" }
}
```

`vite.config.js`
```js
import { defineConfig } from 'vite'
import solid from 'vite-plugin-solid'

export default defineConfig({
  plugins: [solid({ ssr: true })],
  build: { ssr: true, minify: false, target: 'node22' },
})
```

`bench.jsx`
```jsx
import { renderToString } from 'solid-js/web'
import { StaticRouter, Route, A, useResolvedPath, useHref, useLocation } from '@solidjs/router'
import { For } from 'solid-js'

const rows = Array.from({ length: 1000 }, (_, i) => ({ id: `row-${i}`, name: `name-${i}` }))

// Same rendered output as <A>, without the per-instance prop-proxy machinery.
function AFast(props) {
  const inactiveClass = props.inactiveClass ?? 'inactive'
  const activeClass = props.activeClass ?? 'active'
  const to = useResolvedPath(() => props.href)
  const href = useHref(to)
  const location = useLocation()
  const to_ = to()
  let active = false
  let exact = false
  if (to_ !== undefined) {
    const path = to_.split(/[?#]/, 1)[0].replace(/\/$/, '').toLowerCase()
    const loc = decodeURI(location.pathname.replace(/\/$/, '').toLowerCase())
    active = props.end ? path === loc : loc.startsWith(path + '/') || loc === path
    exact = path === loc
  }
  const classes = []
  if (props.class) classes.push(props.class)
  classes.push(active ? activeClass : inactiveClass)
  return (
    <a
      href={href() || props.href}
      state={JSON.stringify(props.state)}
      class={classes.join(' ')}
      link
      aria-current={exact ? 'page' : undefined}
    >
      {props.children}
    </a>
  )
}

function Table(props) {
  return (
    <table>
      <tbody>
        <For each={rows}>
          {(entry) => (
            <tr>
              <td>{entry.id}</td>
              <td>{entry.name}</td>
              <td>
                {props.mode === 'A' ? (
                  <A href={`/item/${entry.id}`}>View</A>
                ) : props.mode === 'AFast' ? (
                  <AFast href={`/item/${entry.id}`}>View</AFast>
                ) : (
                  <a href={`/item/${entry.id}`}>View</a>
                )}
              </td>
            </tr>
          )}
        </For>
      </tbody>
    </table>
  )
}

const render = (mode) =>
  renderToString(() => (
    <StaticRouter url="http://localhost/">
      <Route path="/" component={() => <Table mode={mode} />} />
    </StaticRouter>
  ))

function bench(label, mode) {
  for (let i = 0; i < 30; i++) render(mode)
  const N = 200
  const t0 = performance.now()
  let out
  for (let i = 0; i < N; i++) out = render(mode)
  const ms = (performance.now() - t0) / N
  console.log(
    `${label.padEnd(24)} ${ms.toFixed(2).padStart(7)} ms/render  ${(1000 / ms).toFixed(0).padStart(5)} renders/s  ${out.length} bytes`,
  )
  return { ms, out }
}

const extract = (h) => h.slice(h.indexOf('<table'), h.indexOf('</table>'))
const norm = (h) => h.replaceAll('link="true" ', 'link')

const plain = bench('plain <a> x1000', 'a')
const routerLink = bench('<A> x1000', 'A')
const fast = bench('<A> server fast path', 'AFast')

console.log(
  `\n<A> overhead vs plain <a>: ${(routerLink.ms - plain.ms).toFixed(2)} ms/render (${(routerLink.ms / plain.ms).toFixed(1)}x)`,
)
console.log(
  `fast path vs <A>:          ${(routerLink.ms / fast.ms).toFixed(1)}x faster, output equivalent: ${norm(extract(fast.out)) === norm(extract(routerLink.out))}`,
)
```

`npm install && npm run bench`, Node 26.3.0, Apple silicon:

```
plain <a> x1000             0.31 ms/render   3199 renders/s  133601 bytes
<A> x1000                   6.29 ms/render    159 renders/s  164601 bytes
<A> server fast path        1.12 ms/render    890 renders/s  156601 bytes

<A> overhead vs plain <a>: 5.98 ms/render (20.1x)
fast path vs <A>:          5.6x faster, output equivalent: true
```

`AFast` produces the same markup, modulo `link` serialising as `link` instead of `link="true"` because there is no spread on the element. It is 5.6x faster and 8 bytes per link smaller.

## Profile

`node --cpu-prof dist/bench.js`, self time, 1736 ms active:

```
415 ms  23.9%  mergeProps            solid-js/web/dist/server.js
377 ms  21.7%  (garbage collector)
158 ms   9.1%  split                 solid-js/web/dist/server.js
126 ms   7.2%  splitProps            solid-js/web/dist/server.js
114 ms   6.6%  A                     @solidjs/router
 45 ms   2.6%  children
 43 ms   2.5%  normalizePath
 35 ms   2.0%  ssrElement
```

`mergeProps` plus `splitProps`/`split` is 40% of active CPU, and a large share of the 22% GC is the objects they allocate.

## Root cause

`dist/components.jsx`:

```jsx
export function A(props) {
    props = mergeProps({ inactiveClass: "inactive", activeClass: "active" }, props);
    const [, rest] = splitProps(props, ["href","state","class","activeClass","inactiveClass","end"]);
    ...
}
```

In Solid's server build these are not cheap:

- `mergeProps` runs `Object.getOwnPropertyDescriptors` on each source and installs an `Object.defineProperty` accessor with a closure for **every** key on a fresh object. Every subsequent `props.href` / `props.class` / `props.inactiveClass` read walks the sources array inside that getter.
- `splitProps` then runs `Object.getOwnPropertyDescriptors` again over that all-accessor object, builds two more objects via `defineProperty`, and `delete`s keys out of the descriptor map, which puts it into dictionary mode.

So each `<A>` allocates three accessor-backed objects and pays megamorphic property access on all of them, to emit static markup. It also does `JSON.stringify(props.state)` unconditionally (usually on `undefined`), three `createMemo`s, and a `classList` object spread that is immediately flattened to a string.

None of the reactivity can ever fire during `renderToString`.

## Suggested fix

Give `A` a server implementation that skips `mergeProps`, `splitProps`, and the memos, reading each prop once and emitting the anchor directly, along the lines of the `AFast` above. The router already ships separate server and browser builds, so this can be conditioned on the build rather than a runtime check.

Cheaper partial wins if a full split is not wanted:

- Skip `JSON.stringify(props.state)` when `props.state === undefined`.
- Replace `mergeProps` for the two class defaults with `??` reads at use site.
- Build the class string directly instead of allocating a `classList` object.

## Why this showed up

This surfaced while investigating why SolidStart sits near the bottom of the SSR load benchmark at https://frameworks.e18e.dev (https://github.com/e18e/framework-tracker). Its benchmark page renders 1000 rows each containing an `<A>`. Measured on one machine, same route, same data, autocannon:

| | rps @1 conn | rps @25 conn | p99 @25 conn |
|---|---|---|---|
| SolidStart with `<A>` | 110 | 130 | 678 ms |
| SolidStart with plain `<a>` | 371 | 750 | 67 ms |
| SvelteKit | 329 | 710 | 57 ms |

Replacing `<A>` with `<a>` is the only change, and it takes SolidStart from far behind SvelteKit to level with it. 1000 router links on a page is not typical, so the benchmark amplifies this well past normal usage, but the per-link cost is real and shows up on any list-heavy SSR page.

birkskyum:
Fixes #583.

`<A>` spent most of its server render time in `mergeProps` and `splitProps`, building accessor-backed prop objects that a one shot `renderToString` can never observe. In a 1000 link table the component rendered in **5.89 ms**; it now renders in **1.09 ms**, a **5.4x** improvement, which lands within 10% of a hand written anchor that does no prop plumbing at all.

## What changed

- Drop the `mergeProps` that existed only to default `activeClass` / `inactiveClass`, and read the defaults with `??` at the point of use. Same semantics, since `mergeProps` also only falls back on `undefined`, and the reads stay inside the reactive JSX expression.
- Share the normalized `location.pathname` across links. Every `<A>` on a page ran `normalizePath` + `decodeURI` + `toLowerCase` + a regex on the same string; a one entry cache collapses that to once per navigation. It is a pure function of the input, so a miss can only cost a recompute.
- Skip `JSON.stringify` when there is no `state`. `JSON.stringify(undefined)` returns `undefined`, so the attribute was already omitted.
- Build the `classList` in one object instead of up to three spreads.
- On the server, when the caller passes nothing beyond the props `<A>` consumes itself, skip `splitProps` and the JSX spread entirely. This is the largest single win: the JSX spread compiles to another `mergeProps` per link.

The last one is gated on `isServer` deliberately. A server render is one shot, so the props key set cannot grow after the check. On the client it can, via a reactive spread like `<A {...signal()} />`, and taking the fast path there would silently drop keys added later. I verified that case both ways. Gating also means client builds constant fold the branch away.

## Verification

**Output parity.** 22 prop combinations rendered through `renderToString` (active/inactive, exact and parent matches, `end`, custom `activeClass`/`inactiveClass`, user `class`, user `classList`, `state`, `target`/`rel`, `replace`/`noScroll`/`preload`, `id`/`aria-label`, relative href, query and hash, trailing slash, mixed case, external href, nested children, no children) are byte identical to `main`, except for one insignificant space inside the tag on the fast path, `link="true">` instead of `link="true" >`.

Client DOM output is byte identical on `main`, including the dynamic spread case above.

**Tests.** `<A>` had no test coverage. This adds:

- `test/anchor.spec.tsx`, 15 jsdom tests covering href resolution, active state, `end`, class handling, `classList` merging, state serialization, prop forwarding, children, and active class updates across navigation. All 15 also pass against unmodified `main`, so they are characterization tests rather than tests written to fit the new code.
- `test/ssr/anchor.spec.tsx`, 10 tests covering the server fast path and the spread fallback, plus an assertion that the two paths agree. 8 of these also pass against `main`; the 2 that do not are the ones asserting the exact fast path serialization noted above.

SSR needs the SSR JSX transform and the real `isServer`, so it cannot share the DOM config or its setup file. That is `vitest.ssr.config.ts` and a `test:ssr` script, wired into `pnpm test`.

Full suite: 274 DOM tests, 10 SSR tests, `test:types` clean.

**Size.** Client bundles get slightly smaller, since the server branch folds away and `A` no longer pulls in `mergeProps`. Vite production build importing `A`, `Router`, `Route`:

| | main | this PR |
|---|---|---|
| bundle | 43663 B | 42233 B |
| gzipped | 13485 B | 13117 B |

## Benchmark

Reproduction is in #583. Node 26.3.0, Apple silicon, 1000 links, 200 renders after 30 warmup:

```
                        main      this PR
plain <a> x1000       0.30 ms     0.30 ms
<A> x1000             5.89 ms     1.09 ms
```

Profile before: `mergeProps` 24%, `splitProps`/`split` 16%, GC 22% of active CPU. After, the remaining `<A>` cost is dominated by the work that actually has to happen per link, resolving and normalizing the href.

## Notes

- `link` in the JSX attribute augmentation is widened from `boolean` to `boolean | string`, needed to write `link="true"` on the fast path so its serialization matches what the spread path produces.
- `children` is now in the `splitProps` key list and rendered explicitly, so both paths handle it the same way. Laziness is preserved, the compiler wraps `props.children` in a memo.

### DIFF
--- a/.changeset/spotty-moons-jump.md
+++ b/.changeset/spotty-moons-jump.md
@@ -1,0 +1,16 @@
+---
+"@solidjs/router": patch
+---
+
+Speed up `<A>` server rendering by roughly 5x
+
+`<A>` spent most of its server render time in `mergeProps` and `splitProps`, which build
+accessor-backed prop objects that can never be observed during a one shot
+`renderToString`. The default `activeClass` / `inactiveClass` merge is gone, the
+`location.pathname` normalization is shared instead of repeated per link, `JSON.stringify`
+is skipped when there is no `state`, and on the server a link that passes nothing beyond
+the props `<A>` consumes itself now skips `splitProps` and the JSX spread entirely.
+
+Rendered output is unchanged on the client, and unchanged on the server apart from the
+`link` marker serializing as `link` rather than `link="true"` on that fast path. Client
+bundles get slightly smaller, since the server branch is constant folded away.
--- a/package.json
+++ b/package.json
@@ -31,8 +31,9 @@
   "scripts": {
     "build": "rm -rf dist && tsc && rollup -c",
     "prepublishOnly": "npm run build",
-    "test": "vitest run && npm run test:types",
+    "test": "vitest run && npm run test:ssr && npm run test:types",
     "test:watch": "vitest",
+    "test:ssr": "vitest run --config vitest.ssr.config.ts",
     "test:types": "tsc --project tsconfig.test.json",
     "pretty": "prettier --write \"{src,test}/**/*.{ts,tsx}\"",
     "release": "pnpm build && changeset publish"
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -1,16 +1,9 @@
 /*@refresh skip*/
 import type { JSX } from "solid-js";
-import { createMemo, mergeProps, splitProps } from "solid-js";
-import {
-  useHref,
-  useLocation,
-  useNavigate,
-  useResolvedPath
-} from "./routing.js";
-import type {
-  Location,
-  Navigator
-} from "./types.js";
+import { createMemo, splitProps } from "solid-js";
+import { isServer } from "solid-js/web";
+import { useHref, useLocation, useNavigate, useResolvedPath } from "./routing.js";
+import type { Location, Navigator } from "./types.js";
 import { normalizePath } from "./utils.js";
 
 declare module "solid-js" {
@@ -34,16 +27,55 @@ export interface AnchorProps extends Omit<JSX.AnchorHTMLAttributes<HTMLAnchorEle
   activeClass?: string | undefined;
   end?: boolean | undefined;
 }
+// Every <A> on a page normalizes the same `location.pathname`, so a one entry cache
+// collapses that work to once per navigation instead of once per link. Pure function
+// of the input, so a stale or missed entry can only cost a recompute.
+let lastPathname: string | undefined;
+let lastNormalizedPathname: string;
+function normalizeLocationPath(pathname: string): string {
+  if (pathname !== lastPathname) {
+    lastNormalizedPathname = decodeURI(normalizePath(pathname).toLowerCase().replace(/\/$/, ""));
+    lastPathname = pathname;
+  }
+  return lastNormalizedPathname;
+}
+
+// Props <A> consumes itself. `children` is deliberately absent: leaving it in `rest` keeps
+// the spread path byte for byte identical to before, including `innerHTML`/`textContent`
+// precedence over children and the order a `ref` observes.
+const SPLIT_PROPS = [
+  "href",
+  "state",
+  "class",
+  "activeClass",
+  "inactiveClass",
+  "end"
+] as const satisfies readonly (keyof AnchorProps)[];
+
+// The fast path renders children explicitly, so a link that only has children still
+// qualifies for it. Anything outside this set, `innerHTML` included, takes the spread path.
+const FAST_PATH_PROPS: ReadonlySet<string> = new Set([...SPLIT_PROPS, "children"]);
+
 export function A(props: AnchorProps) {
-  props = mergeProps({ inactiveClass: "inactive", activeClass: "active" }, props);
-  const [, rest] = splitProps(props, [
-    "href",
-    "state",
-    "class",
-    "activeClass",
-    "inactiveClass",
-    "end"
-  ]);
+  // `splitProps` plus the JSX spread is the bulk of the per link cost, and neither is
+  // needed when the caller passes nothing beyond the props <A> consumes itself. Server
+  // only: a render pass there is one shot, so the key set cannot grow after this check,
+  // which it can on the client when the caller uses a reactive spread. The branch is
+  // constant folded out of client builds.
+  //
+  // Own property names rather than `for...in`: a non-enumerable own prop still reaches the
+  // element through `splitProps`, so missing one here would silently drop it.
+  let fastPath = false;
+  if (isServer) {
+    fastPath = true;
+    for (const key of Object.getOwnPropertyNames(props)) {
+      if (!FAST_PATH_PROPS.has(key)) {
+        fastPath = false;
+        break;
+      }
+    }
+  }
+
   const to = useResolvedPath(() => props.href);
   const href = useHref(to);
   const location = useLocation();
@@ -52,19 +84,44 @@ export function A(props: AnchorProps) {
     if (to_ === undefined) return [false, false];
     // trailing slashes are ignored so `/route` and `/route/` share active state
     const path = normalizePath(to_.split(/[?#]/, 1)[0]).toLowerCase().replace(/\/$/, "");
-    const loc = decodeURI(normalizePath(location.pathname).toLowerCase().replace(/\/$/, ""));
+    const loc = normalizeLocationPath(location.pathname);
     return [props.end ? path === loc : loc.startsWith(path + "/") || loc === path, path === loc];
   });
+  // One read, so a getter backed `state` is not observed twice
+  const state = () => {
+    const value = props.state;
+    return value === undefined ? undefined : JSON.stringify(value);
+  };
+
+  if (fastPath) {
+    return (
+      <a
+        href={href() || props.href}
+        state={state()}
+        classList={{
+          ...(props.class && { [props.class]: true }),
+          [props.inactiveClass ?? "inactive"]: !isActive()[0],
+          [props.activeClass ?? "active"]: isActive()[0]
+        }}
+        link
+        aria-current={isActive()[1] ? "page" : undefined}
+      >
+        {props.children}
+      </a>
+    );
+  }
+
+  const [, rest] = splitProps(props, [...SPLIT_PROPS]);
 
   return (
     <a
       {...rest}
       href={href() || props.href}
-      state={JSON.stringify(props.state)}
+      state={state()}
       classList={{
         ...(props.class && { [props.class]: true }),
-        [props.inactiveClass!]: !isActive()[0],
-        [props.activeClass!]: isActive()[0],
+        [props.inactiveClass ?? "inactive"]: !isActive()[0],
+        [props.activeClass ?? "active"]: isActive()[0],
         ...rest.classList
       }}
       link
--- a/test/anchor.spec.tsx
+++ b/test/anchor.spec.tsx
@@ -1,0 +1,221 @@
+// @vitest-environment jsdom
+import { vi } from "vitest";
+import { render } from "solid-js/web";
+import { A, MemoryRouter, Route, createMemoryHistory, useNavigate } from "../src/index.jsx";
+
+// jsdom has no scrollTo, and navigating triggers the router's scroll handling
+window.scrollTo = vi.fn() as any;
+
+function mount(url: string, Comp: () => any) {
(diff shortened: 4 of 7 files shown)


birkskyum:
Pushed a revision. An adversarial review of the first version turned up four behavior differences from `main` that I had missed. All four are fixed, and each now has a regression test that fails on the previous commit and passes on both `main` and this branch.

### 1. `children` broke `innerHTML` / `textContent` and `ref` ordering

Moving `children` into the `splitProps` list meant the fallback always supplied an explicit children expression, so `ssrElement` ignored child-like props from `rest`:

| | `main` | previous commit |
|---|---|---|
| `<A innerHTML="<b>inside</b>" />` | `<b>inside</b>` | empty anchor |
| `<A textContent="inside" />` | `inside` | empty anchor |
| `innerHTML` + children | `innerHTML` wins | children won |

On the client a `ref` callback saw `""` instead of `"child"`, because the spread processed `ref` before the explicit insertion.

Fixed by splitting the key sets. `children` stays out of `splitProps` and the fallback is a self-closing `<a {...rest} ... />` exactly as before; `children` is only in the fast path whitelist and only rendered explicitly there, where `innerHTML` cannot occur by construction.

### 2. `state` was read twice

The `!== undefined` check and the `JSON.stringify` were separate reads, so a getter-backed `state` was observed twice. Now read once into a local.

### 3. The class map mishandled `__proto__`

Building the map with `list[key] = value` and `Object.assign` uses `[[Set]]`, which hits the `__proto__` setter and reassigns the prototype instead of defining an own property. `class="__proto__"`, `inactiveClass="__proto__"`, and `classList={{["__proto__"]: true}}` all silently lost the class.

Reverted to the original object-literal form, which uses `[[DefineOwnProperty]]`. This cost nothing measurable, so it was never worth the risk.

### 4. The fast path check missed non-enumerable own props

`for...in` skips non-enumerable own properties, but `splitProps` still forwards them, so `createComponent(A, props)` with a non-enumerable `id` silently dropped it. Now uses `Object.getOwnPropertyNames`, which is the conservative direction: anything unrecognized falls back to the spread.

### Also changed

Reverted the `link?: boolean` → `boolean | string` widening. Widening a global anchor attribute type just to preserve exact serialization was not a good trade, so the fast path emits bare `link`. The router only ever does `hasAttribute("link")`. This is the one remaining output difference from `main`, and it also drops ~7 KB from a 1000-link page.

Tests no longer assert exact whitespace or the `link` representation, and the navigation test uses `vi.waitFor` instead of a fixed delay.

### Verification

- **30 differential SSR cases** (the original 22 plus 8 adversarial ones) now render identically to `main`, modulo the `link` marker form.
- **Client DOM output** identical to `main`, including the dynamic-spread case.
- Suite: 275 DOM tests, 18 SSR tests, types clean.
- The 9 new regression tests fail on the previous commit; all 18 SSR tests and all 16 DOM tests pass unmodified against `main`.

**Benchmark**, three alternating build rounds to control for drift:

| round | `main` | this branch | speedup |
|---|---|---|---|
| 1 | 5.90 ms | 1.11 ms | 5.32x |
| 2 | 5.93 ms | 1.13 ms | 5.25x |
| 3 | 5.81 ms | 1.13 ms | 5.14x |

**Client bundle**, clean-room Vite production build importing `A`, `Router`, `Route`:

| | `main` | this branch |
|---|---|---|
| raw | 43663 B | 42151 B |
| gzipped | 13485 B | 13081 B |

I confirmed the server branch is fully constant folded: the client bundle contains a single `<a>` template and no trace of the fast path check.

### Known remaining headroom, not in this PR

The fast path only helps links with no extra props. A link with any extra prop still costs ~3.9 ms per 1000, since it pays both `splitProps` and the spread's `mergeProps`. Building a single server anchor-props object and emitting one spread gets that to ~1.1 ms, and a descriptor-preserving variant to ~1.4 ms. It needs careful precedence handling for `children`, `innerHTML`, `classList`, `link`, and `aria-current`, so I have left it out rather than grow this PR. Happy to follow up if you want it.

Two smaller ones I measured and rejected: bypassing the three server `createMemo`s via direct route resolution is worth 4-5% and not worth the extra branching, and replacing `split(/[?#]/)` with `search`/`slice` made no measurable difference.

PR #590, opened 2026-08-27
Adopt path-based server function addressing

ryansolid:
## Summary

Downstream adoption of [solidjs/solid#3076](https://github.com/solidjs/solid/pull/3076), which moved the server function id into the path (`<endpoint>/<id>`) and retired both the `X-Server-Function-Id` header and the `?id=` query fallback.

- The generic form-action fallback (`createServerFormAction`) reads the id back through the runtime's new `parseServerFunctionUrl` instead of parsing `?id=` by hand — the router no longer hard-codes the addressing scheme.
- Bound `?args` composition (`action.with`) is untouched: arguments stayed in the query per the [#3072 discussion](https://github.com/solidjs/solid/discussions/3072), and `hashKey` already emits the plain JSON the server-side convention decodes.
- "Not an address" now means a malformed path (bare endpoint, extra segments) rather than a missing `?id`; the native-submission fallbacks keep working through the runtime's parse answering `null`.
- `actionBase` prefix interception is unchanged — `/_server/<id>` still lives under `/_server`.

## Draft until

`@solidjs/web` publishes the addressing change (anything newer than `2.0.0-rc.3`). The `^2.0.0-rc.3` range picks the new build up on publish; CI here will fail on the missing `parseServerFunctionUrl` export until then.

## Test plan

- [x] Full suite (dom + server configs + type tests) passes locally against a `file:` link to the `@solidjs/web` build from solid `next` — 382 + 33 passing.
- [ ] Re-run CI once the next `@solidjs/web` prerelease is published.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/adopt-path-based-server-function-addressing.md
+++ b/.changeset/adopt-path-based-server-function-addressing.md
@@ -2,4 +2,4 @@
 "@solidjs/router": patch
 ---
 
-Adopt path-based server function addressing (solidjs/solid#3076). Action urls are now `<endpoint>/<id>[?args=...]` — the id in the path, bound arguments staying in the query. The generic form-action fallback reads the id back through the runtime's `parseServerFunctionUrl` instead of parsing `?id=` by hand, so the router no longer hard-codes the addressing scheme. Requires `@solidjs/web` newer than 2.0.0-rc.3.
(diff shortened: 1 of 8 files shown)


ryansolid:
Companion plugin PR: solidjs/solid-vite-plugin#332 (endpoint prefix-matching + path-based module preload in the dev middleware). Both are draft until the next `@solidjs/web` prerelease publishes the addressing change.

PR #609, opened 2026-09-18
Follow @solidjs/web rc.9: parseServerFunctionActionUrl

ryansolid:
Lockstep follow-up for `@solidjs/web` 2.0.0-rc.9.

solidjs/solid#3440 renamed the form-post address parser: `parseServerFunctionUrl` → `parseServerFunctionActionUrl` (no alias kept), with `serverFunctionUrl(fn, ...args)` now naming the url a `GET()` reference's own call requests. The router imports the old name in `createServerFormAction` (`src/data/action.ts`), so on rc.9 the module fails to link.

- `src/data/action.ts`: rename the import and the call. Same url shape, same id — no behavior change.
- `package.json`: `solid-js` / `@solidjs/web` dev + peer floor → `^2.0.0-rc.9`.
- changeset (patch).

Draft until rc.9 is on npm — the lockfile can't be updated before then, so CI's `--frozen-lockfile` install will fail until I push the lockfile bump. Everything else in `next` already resolves against rc.9: no `INSTANCE_HEADER`, `context.instance`, `observeServerFunctionCalls`, `<Dynamic>`, or `solid-js/internal` moves in use; the flight consumer already takes `data` as optional, so solidjs/solid#3502's `undefined` delivery is fine as-is (and #608 handles `X-Revalidate: *`).

Made with [Cursor](https://cursor.com)


ryansolid:
> rc.9 is on npm; lockfile pushed. Typecheck is clean, but the suite is **12 failed / 390 passed** against rc.9, where `origin/next` on rc.8 is 401/401 — so this PR is not mergeable as-is, and the rename is the smallest part of the work.
>
> Only one failure is expectation drift: `generic server actions › synthesizes an action for an unregistered server-action url` asserts `X-Server-Function-Instance` is set; rc.9 removed the instance header (`INSTANCE_HEADER` / `context.instance`), so that line should flip to `toBeUndefined()`.
>
> The other 11 are behavioral, none of them click-driven, all around navigation commit under transitions:
>
> - `navigate-commit.spec.tsx` (3): superseding redirects commit with the wrong history policy; navigation intent not derived from the pending source transition; rapid-fire bursts land on `/b` instead of `/a`.
> - `redirect-retained.spec.tsx` (3): the leaving section's eager refetch runs (`2` fetches, expected `1`) in all three.
> - `router.spec.ts` (2): `isRouting` stays `false` when a push causes a transition; the 100-navigations-in-one-update guard no longer throws.
> - `search-params.spec.tsx` (2): two synchronous `setSearchParams` calls — the first is lost; `setSearchParams` during a pending navigation applies to `/`, not the target.
> - `action.spec.ts` (1): `onSubmit` writing to a `createOptimistic` primitive — the action function reads `initial`, not the optimistic value.
>
> These share a shape: work that used to ride the pending navigation's transition/optimistic lane is now landing outside it. That's a core change between rc.8 and rc.9 the router needs to either adapt to or report as a regression — it isn't resolved here. Leaving this as a draft with the rename + lockfile until that's sorted.
>
> — *Claude via Cursor*

ryansolid:
> Correction to the above: these are not core regressions. All 12 trace to deliberate rc.9 rules, and the router is written against the rc.8 semantics.
>
> **A28** (`solidjs/solid@8ff4803`, "a write becomes visible at flush — to every channel"): between `set(x)` and the flush, `latest(x)` answers the pre-write value and `isPending(x)` is `false`; it explicitly supersedes the #2922 mid-tick `latest()` pull. The router's `pendingNavigation()` is exactly that pull — `isRouting()` → `isPending(source)`, `pendingTarget` → `latest(source)`, both read synchronously after `setSource(...)`. That accounts for 10 of the 12: both `setSearchParams` cases (merge target is `undefined` mid-tick), `isRouting` after a push, the 100-navigation guard, and the six `navigate-commit` / `redirect-retained` cases (hop detection is `isPending(source)` at `routing.ts:1052`, so a hop reads as a fresh navigation).
>
> The `onSubmit` → `createOptimistic` case is A28 (5): `setOptimistic(v)` becomes the override at the flush that carries it; plain reads see nothing before. The `X-Server-Function-Instance` assertion is the header rc.9 removed on purpose.
>
> Adaptation: keep the in-flight target in the router's own state (every location write already funnels through the one integration `write` in `factory.tsx`) instead of pulling it back out of the signal mid-tick; flip the instance-header assertion to `toBeUndefined()`.
>
> — *Claude via Cursor*

ryansolid:
> Second correction — the adaptation I proposed above (router owns the in-flight target) is wrong too. `navigate()` is `setSource(...)`, a plain signal write, so under A28 the router isn't special: the write turns pending/`latest` at the flush that carries it, like every other signal, and the router shouldn't flush eagerly or defer to work around that.
>
> What A28 actually asks of a writer that needs to compose synchronously is its own channel: *"a functional updater, the store draft … compose on it."* So the split is:
>
> - **Router code:** `setSearchParams` composes by reading back `router.pendingTarget` (`latest(source)` of its own unflushed write). It should compose through the updater — `setSource(prev => …)` — and the second synchronous call then sees the first without a flush. Same for anything in `navigateFromRoute` that reads `latest(source)` to build the next write.
> - **Tests:** `isRouting()` true immediately after `navigate()`, `pendingTarget` populated mid-tick, the 100-navigations guard firing inside one synchronous burst — these assert the rc.8 mid-tick pull that `solidjs/solid@8ff4803` supersedes. They flush before asserting, as core's own tests were re-pinned. The `X-Server-Function-Instance` assertion flips to `toBeUndefined()`.
>
> Hop detection (`routing.ts:1052`) reads `isPending(source)` from a guard that runs during the landing flush, so it's already on the right side of the rule; I expect most of the six `navigate-commit` / `redirect-retained` cases to be test-side as well, to be confirmed after the updater change.
>
> — *Claude via Cursor*

